PR #9, opened 2022-01-26
Basic Ingestion Function

NatHillardUSDS:
Creates basic ingestion function https://app.clickup.com/t/22w34xr


.funcignore
```diff
@@ -0,0 +1,5 @@
+.git*
+.vscode
+local.settings.json
+test
+.venv
```
JSkeets, line 5: nitpick: newline

.gitignore
```diff
@@ -1 +1,135 @@
...
+bin
+obj
+appsettings.json
+local.settings.json
+
+# Azurite artifacts
+__blobstorage__
+__queuestorage__
+__azurite_db*__.json
+.python_packages
```
JSkeets, line 135: nitpick: newline

.vscode/settings.json
```diff
@@ -1,6 +1,6 @@
 {
   "cSpell.words": ["ATSDR", "CFPB", "ISSO", "cybersecurity"],
```
JSkeets, line 2: Do we want these vscode files checked into source control ?

src/FunctionApps/PITest_FunctionApp/IngestFunction/sample.dat
```diff
@@ -0,0 +1,3 @@
+{
+    "name": "Azure"
+}
```
JSkeets, line 3: nitpick: newline

src/FunctionApps/PITest_FunctionApp/host.json
```diff
@@ -0,0 +1,15 @@
...
+                "isEnabled": true,
+                "excludedTypes": "Request"
+            }
+        }
+    },
+    "extensionBundle": {
+        "id": "Microsoft.Azure.Functions.ExtensionBundle",
+        "version": "[3.3.0, 4.0.0)"
+    }
+}
```
JSkeets, line 15: nitpick: newline

docs/getting_started.md
```diff
@@ -0,0 +1,5 @@
+# Developer Getting Started Guide
```
JSkeets, line 1: I think I would prefer to consolidate documentation in one place. I.e. it is all in the README. Just my 2 cents.

src/FunctionApps/PITest_FunctionApp/IngestFunction/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    file_object = BytesIO()
+    sftp.getfo(target_file_path, file_object)
+    file_object.seek(0)
+    logging.debug(
+        f"Uploading {target_file_name} at path {target_file_path} Azure Storage"
+    )
+    upload_blob(target_file_name, file_object)
+
+
+def main(req: func.HttpRequest) -> func.HttpResponse:
```
JSkeets, line 74: I am actually having a hard time tracking how this gets triggered to completion.
1) This function gets called and it creates the sftp connection
2) Print tree gets called...
3) I don't see where `test_decryption` is called which would then call `upload_blob` 

This is likely just me needing an explanation which I greatly appreciate.

workspace.code-workspace
```diff
@@ -0,0 +1,11 @@
...
+    "folders": [
+		{
+			"path": "src/FunctionApps/GeocodeAddress"
+		},
+		{
+			"path": "src/FunctionApps/PITest_FunctionApp"
+		}
+	],
+    "settings": {}
+}
```
JSkeets, line 11: nitpick: Newline

JSkeets:
Couple outstanding questions, which are in the comments. Has this been tested ? I am hoping that we can generate a culture of providing proof of testing in the PR via screenshots or whatever means possible.

NatHillardUSDS:
Thank you @JSkeets for the comments! 100% agreed on all fronts. Just now I was looking into testing Azure blob storage, I'm planning to push up a `tests` folder shortly that uses [Azurite](https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio) to do this!

.funcignore
NatHillardUSDS: Ah right - these files were auto-generated by VSCode but agreed we should enforce EOF new lines in general (VSCode also creates `^M`s at the end of each line that I would rather do without...). I can add an explicit `Black` config here which would ensure these are present

.vscode/settings.json
NatHillardUSDS: I'm not sure where this particular list came from - I can look into this - in general though these enable local development and testing and the Azure extension in VSCode intentionally does not exclude them from the built product. I can add a `README` about how to run these locally, and meanwhile I think I can just remove this particular line, which I think was added here spuriously. The rest of this file is important for local development

docs/getting_started.md
NatHillardUSDS: This makes sense - this was originally modeled after what ReportStream does, but agreed, I'm not sure it adds much value here for them to be separate

src/FunctionApps/PITest_FunctionApp/IngestFunction/__init__.py
NatHillardUSDS: Ah, thanks for catching this - I need to remove the `test_decryption` method, as this was just meant for that single file; the solution really should attempt to run `get_r` against the dir tree to copy everything over . I had this change on a separate branch but neglected to remove this here

src/FunctionApps/PITest_FunctionApp/IngestFunction/settings.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+
+class Settings:
+    hostname = os.environ.get("VDHSFTPHostname")
```
jeremyzitomer-usds, line 5: where can we find documentation on where these are set? i presume in some function configuration on azure? are these variables initially created via terraform?

src/FunctionApps/PITest_FunctionApp/IngestFunction/settings.py
```diff
@@ -0,0 +1,11 @@
...
+    username = os.environ.get("VDHSFTPUsername")
+    password = os.environ.get("VDHSFTPPassword")
```
jeremyzitomer-usds, line 7: nit: can we use the same NAMING_CONVENTION for all env vars?

src/FunctionApps/PITest_FunctionApp/IngestFunction/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    logging.info("Python HTTP trigger function processed a request.")
```
jeremyzitomer-usds, line 75: i feel like this should say "received a request" -- the subsequent logic below this is what is required to "process" the request, no?

src/FunctionApps/PITest_FunctionApp/IngestFunction/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    logging.info(f"Settings: {settings}")
```
jeremyzitomer-usds, line 76: careful -- this won't print out our VA SFTP username and password in plaintext, will it???

src/FunctionApps/PITest_FunctionApp/IngestFunction/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+    try:
+        cnopts = pysftp.CnOpts()
+        cnopts.hostkeys = None
+        sftp = pysftp.Connection(
+            settings.hostname,
+            username=settings.username,
+            password=settings.password,
+            cnopts=cnopts,
+        )
+        logging.info("Top level directory listing:")
```
jeremyzitomer-usds, line 87: the output of this isn't sensitive, is it? are the directory/file names etc themselves revealing at all?

src/FunctionApps/PITest_FunctionApp/IngestFunction/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+        # for file_path in sftp.listdir("/ELR"):
```
jeremyzitomer-usds, line 88: is there a compelling reason to leave this comment in?

src/FunctionApps/PITest_FunctionApp/IngestFunction/__init__.py
jeremyzitomer-usds: also, is there anything in the content of the http request we should be logging?

src/FunctionApps/PITest_FunctionApp/IngestFunction/__init__.py
jeremyzitomer-usds: Still, a high-level comment on the `main` function here would be nice

jeremyzitomer-usds (commented):
Pardon the late night review :) my brain is operating at odd hours this weekend.

also, nit: if we have control over the name at this stage, can we call PITest_FunctionApp something that doesn't have test in the name? i know we mentioned this the other day, but hopefully at this level we have some leeway?

.vscode/settings.json
bamader: I would be as well

docs/getting_started.md
bamader: Major preference for single stop documentation. To me splitting it out only makes it harder to actually _get_ started.

src/FunctionApps/PITest_FunctionApp/IngestFunction/settings.py
bamader: ^^^ 💯  for this please

src/FunctionApps/PITest_FunctionApp/IngestFunction/__init__.py
bamader: So the action items I'm seeing here are:
1. delete the `test_decryption` function in its entirety.
2. Make a call _somewhere_ to `get_r` to recursively copy the directory structure from remote to local.

Does this also imply that we're dumping `upload_blob`, or is that something we'll call after we've copied the directories over?

bamader (approved):
Good to me after the `test_decryption` / `get_r` code is sorted out (pending a few nits and consolidations).

PR #13, opened 2022-01-28
Josiahsiegel/storage hadr

JosiahSiegel:
* Update storage HADR
* Add monitoring to storage
* Clarify resource group name variable


cwinters-usds (approved):
:shipit:

PR #14, opened 2022-01-28
adding metric change that Azure forced behind the scenes

JosiahSiegel:
Including change made by Azure behind the scenes to keep TF clean


JosiahSiegel:
This metric did not show as an option on the frontend, so I'm not going to include any changes other than what Azure did on the backend to avoid issues.

cwinters-usds (approved):
:shrug:

PR #16, opened 2022-02-01
Update .gitignore

NatHillardUSDS:
Removing vscode reference from gitignore - we actually probably want this to share vscode configs


bryan-skylight (approved):
What's better than getting impact by reviewing a single diff? Getting impact by reviewing TWO diffs because your first review missed something!

PR #18, opened 2022-02-01
Add Azure Function to Parse and Validate Data from VDH

bryan-skylight:


JSkeets:
@bryan-skylight We need to make a PR template for the team, but would you mind providing a brief write up of your PR. At a minimum I think it should contain the following:

# Description

# Screenshots (if applicable)

# Any specific feedback you desire ( if applicable)

bryan-skylight:
# Description
This PR moves the code for the Azure Functions I developed from my local development into our prod environment. The code is being moved in its entirety inside of the FunctionApps directory that already exists in this project. The goal of this PR is to add this functionality to our cloud environment to extend our workflow such that it now looks like this:

```VDH uploads data to SFPT 
        --> HTTP Trigger is hit in Azure, which pulls the data from SFTP and stores it in its raw form in a blob 
        --> Blob Trigger sees the data, decrypts it, and stores it in a new blob 
        --> ParseHL7.kt (from this PR) sees the decrypted data via a Blob Trigger and it reads, parses, and validates the data
        --> Valid data are stored in one blog storage while invalid data are stored in another
```

# Screenshots
N/A - I suppose I could send a screenshot of the data existing in the test storage containers I created, but now sure that proves this functionality works. I might add a video recording of the process if we think that's worthwhile.

# Desired Feedback
1. The routes to the storage containers that I'm using (i.e. `received-files`, `valid-messages`, and `invalid-messages`) don't exist in our prod environment, so we'll need to change those out. I assume that in an ideal world, this PR wouldn't be created until those paths have been updated, but we wanted to push the code now so that Chris Winters and his team would have something to reference when we asked them to set up the infra for us in Terraform. So, one thing I'd like to know is what we think we should do in situations like this. Should we use placeholder names? Should we determine what the bucket names are/will be first?
2. We currently have three functions under the `FunctionApps` directory: `GeocodeAddress`, which is a Python function, and `InjestHL7` & `ParseHL7`, which are both Java apps. Do we see an issue with this? Does anything special need to be done with our folder structure, or the code that's being pushed to make this work?
3. Do we want to keep the `InjestHL7.kt` function? It allows users to upload data to a storage container, but our current workflow requires the use of SFTP. If we're not going to use it, I'll delete it.
4. Does everyone agree with how I've broken down the code. Everything that currently lives in the `Utilities` subdirectory could potentially be pulled out, made as its own package and made publicly available so that our Azure Functions merely needed to import it. But more importantly, do we agree that breaking the functionality down into Reader, Parser, and Validator classes makes sense?

NatHillardUSDS:
> @bryan-skylight We need to make a PR template for the team, but would you mind providing a brief write up of your PR. At a minimum I think it should contain the following:
> 
> # Description
> # Screenshots (if applicable)
> # Any specific feedback you desire ( if applicable)

I like this template! We can also add it in `.github/pull_request_template.md` for future use as documented [here](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository).

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/Utilities/HL7Validator.kt
```diff
@@ -0,0 +1,26 @@
...
+        val context = DefaultHapiContext()
+        // We will likely replace this with a more custom validation class in the future
+        context.setValidationContext(ValidationContextFactory.defaultValidation() as ValidationContext);
+        val parser = context.getPipeParser()
+
+        try {
+            parser.parse(message)
+            return true
+        } catch (e: HL7Exception) {
+            return false
```
JSkeets, line 23: Should there be logging here ? It looks to me like this would fail silently ?

bryan-skylight: For this very basic approach that we're using right now, I think failing silently is okay. The gist of this snippet is to try to parse the HL7 message using the default validator, and if it fails for any reason related to HL7 then the message is deemed invalid. This does raise two points: 1) do we want to log _why_ the message failed and use that to inform the sender, and 2) do we need to incorporate a second `catch` condition that grabs all other exceptions (those not related to HL7)?

NatHillardUSDS (changes requested):
Overall looks good - a few comments to address; overall also I would wonder if its possible to add simple unit tests showing what invalid vs. valid might mean

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/InjestHL7.kt
```diff
@@ -0,0 +1,39 @@
...
+import java.nio.charset.StandardCharsets
+
+class HL7Ingest {
+    @FunctionName("injesthl7")
+    fun run(
+        @HttpTrigger(
+            name = "req",
+            methods = [HttpMethod.GET, HttpMethod.POST],
+            authLevel = AuthorizationLevel.FUNCTION
+        ) request: HttpRequestMessage<Optional<String>>,
```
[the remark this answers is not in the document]
bryan-skylight: This method isn't included in the PR anymore. This was originally in there as a means to mock your decrypt function, but given the decrypt function, I've removed it to avoid confusion. This applies to your comment below as well.

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/Utilities/HL7Parser.kt
```diff
@@ -0,0 +1,73 @@
...
+        return processedMessages
+    }
+
+    public fun convertBatchMessagesToList(
+        content: String,
+        delimiter: String = "\n"
+    ): MutableList<String> {
+        var reg = "[\r\n]".toRegex()
+        var cleanedMessage: String = reg.replace(content, delimiter)
+        reg = "[\\u000b\\u001c]".toRegex()
```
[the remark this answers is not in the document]
bryan-skylight: Done!

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/Utilities/HL7Parser.kt
```diff
@@ -0,0 +1,73 @@
...
+                val parsedMessage = parser.parse(it)
+                processedMessages["valid_messages"]!!.add(parsedMessage)
+            } catch (e: HL7Exception) {
+                processedMessages["invalid_messages"]!!.add(it)
+            }
+        }
+        return processedMessages
+    }
+
+    public fun convertBatchMessagesToList(
```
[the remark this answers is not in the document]
bryan-skylight: Done!

src/FunctionApps/DSTP/extensions.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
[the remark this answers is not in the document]
bryan-skylight: Done!

src/FunctionApps/DSTP/.gitignore
```diff
@@ -0,0 +1,36 @@
...
+
+# IDE
+.idea/
+*.iml
+
+# macOS
+.DS_Store
+
+# Azure Functions
+local.settings.json
```
jeremyzitomer-usds, line 33: we should definitely find some way to add this to a top level .gitignore -- this is too sensitive a file to rely on us to remember to gitignore it for each individual function app

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/ParseHL7.kt
```diff
@@ -0,0 +1,66 @@
...
+        ) validContent: OutputBinding<String>,
+        @BlobOutput(
+            name="invalidTarget",
+            dataType = "string",
+            path="bronze/validation/{name}.invalid",
+            connection="AzureWebJobsStorage"
+        ) invalidContent: OutputBinding<String>,
+        context: ExecutionContext
+	) {
+        context.logger.info("Name: ${filename} Size: ${content.size} bytes.")
```
jeremyzitomer-usds, line 35: is the filename ever sensitive?

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/Utilities/HL7Reader.kt
```diff
@@ -0,0 +1,36 @@
...
+import java.io.InputStream
+import java.io.OutputStream
+import java.nio.charset.StandardCharsets
+
+public class HL7Reader() {
+    fun readHL7MessagesFromFile(
+        filepath: String
+    ): String {
+        // reads in a file of one or more HL7 messages and returns the content as a string
+        return "This method is not currently implemented."
```
jeremyzitomer-usds, line 14: this should raise an error, not return a dummy string, no? ditto below.

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/Utilities/HL7Reader.kt
```diff
@@ -0,0 +1,36 @@
...
+        filepath: String 
+    ): String {
+        // given a cloud client, bucket name, and file path, read the content and return it as a string
+        return "This method is not currently implemented."
+    }
+
+    fun readHL7MessagesFromByteArray(
+        content: ByteArray
+    ): String {
+        val rawMessage: String = String(content, StandardCharsets.UTF_8)
```
jeremyzitomer-usds, line 29: is this worth commenting to explain what's happening here? or is this just super straightforward byte conversion

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/ParseHL7.kt
```diff
@@ -0,0 +1,66 @@
...
+
+        // we don't actually do any processing of the HL7 messages here
+        // We read in files from blob storage, extract individual messages from that
+        // file, and then check that each message is valid. If it is, it goes to one
+        // bucket, and if it's not it goes to another.
+
+        val reader = HL7Reader()
+        val parser = HL7Parser()
+        val validator = HL7Validator()
+        val cleanedMessages: String = reader.readHL7MessagesFromByteArray(content)
```
jeremyzitomer-usds, line 45: "cleaned" implies that there's some processing other than byte conversion here -- can we find another variable name?

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/Utilities/HL7Parser.kt
```diff
@@ -0,0 +1,80 @@
...
+           end of the message.
+        */
+        reg = "[\\u000b\\u001c]".toRegex()
+        cleanedMessage = reg.replace(cleanedMessage, "").trim()
+        val messageLines = cleanedMessage.split(delimiter)
+        val nextMessage = StringBuilder()
+        val output = mutableListOf<String>()
+        
+        messageLines.forEach {
+            if (it.startsWith("FHS"))
```
jeremyzitomer-usds, line 56: please provide a comment on the significance of these prefixes

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/Utilities/HL7Parser.kt
```diff
@@ -0,0 +1,80 @@
...
+                processedMessages["invalid_messages"]!!.add(it)
+            }
+        }
+        return processedMessages
+    }
+
+    /* This method was adopted from PRIME ReportStream, which can be found here: 
+       https://github.com/CDCgov/prime-reportstream/blob/194396582be02fcc51295089f20b0c2b90e7c830/prime-router/src/main/kotlin/serializers/Hl7Serializer.kt#L121
+    */
+    public fun convertBatchMessagesToList(
```
jeremyzitomer-usds, line 39: just a one sentence comment on what this is doing would be nice, even though it's copied from RS.

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/Utilities/HL7Validator.kt
jeremyzitomer-usds: if there's some way to do #1 that does not expose sensitive data, yes pls. and #2 yes but probably one level up from this function?

jeremyzitomer-usds (changes requested):
forgive the late night review, brain working at odd hours this weekend :)

i generally like the way you've broken it down here. would be nice to have some top-level exception-catching mechanism, though. otherwise this could fail silently or fail partway through, etc.

also perhaps I'm missing something, but doesn't a given function app have to consist of functions of one and only one language? ie if we want kotlin functions alongside python functions, we need to create multiple apps?

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/Utilities/HL7Reader.kt
bryan-skylight: This is just straightforward byte conversion

bryan-skylight:
Force pushed an update to undo the accidental merge I had created. This cleans up the PR.

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/ParseHL7.kt
bryan-skylight: I'm not sure, so to err on the side of caution, I've commented out the logger functionality. We can reassess at a later date.

NatHillardUSDS (approved):
LGTM!

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/Utilities/HL7Validator.kt
bryan-skylight: I'm going to add logging and error in the next iteration of this code as the process is going to change slightly. I'll address these concerns in that PR. Resolving this comment for now.

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/Utilities/HL7Parser.kt
```diff
@@ -0,0 +1,117 @@
...
+import ca.uhn.hl7v2.validation.ValidationContext;
+import ca.uhn.hl7v2.validation.impl.ValidationContextFactory;
+
+public class HL7Parser() {
+    data class ProcessedMessages(
+        val valid_messages: List<Message>,
+        val invalid_messages: List<String>
+    )
+
+    public fun parse(content: String): ProcessedMessages {
```
graham-skylight, line 18: Is this currently being called?

graham-skylight (commented):
Looks good overall - couple pretty minor things:

* `"parse"` is a little confusing here - we've got a `ParseHL7` class, an `HL7Parser` class, and then `ca.uhn.hl7v2.parser.Parser` from hapi
* I don't know a ton about idiomatic kotlin, but I wonder if we could use functions instead of classes for the main building blocks here (afaict they're stateless)
* It seems pretty likely that we're not going to use a *ton* of kotlin on this project, but if it's an area we're likely to work on we could probably benefit from some basic linting/formatting/etc

src/FunctionApps/DSTP/pom.xml
```diff
@@ -0,0 +1,301 @@
...
+            <version>${kotlin.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <version>5.4.2</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
```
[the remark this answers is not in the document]
bryan-skylight: We're not doing any mocking yet as we're not testing the Azure functions themselves. Testing the functionality that's being called by the Azure functions has been doable thus far without that need.

src/FunctionApps/DSTP/src/main/kotlin/gov/cdc/prime/phdi/Utilities/HL7Parser.kt
bryan-skylight: It is not currently being called. It's a placeholder for anticipated future functionality.

graham-skylight:
sure - neither are huge points, just maybe something to think about down the line:

* functions vs classes: I think it'd be pretty trivial to wrap a function in a class should we need to
* linting/formatting: looks like ktlint does a decent lint and autoformat (https://ktlint.github.io/) but I'm not sure if that's the best

bryan-skylight:
> sure - neither are huge points, just maybe something to think about down the line:
> 
> * functions vs classes: I think it'd be pretty trivial to wrap a function in a class should we need to
> * linting/formatting: looks like ktlint does a decent lint and autoformat (https://ktlint.github.io/) but I'm not sure if that's the best

I'm going to merge this PR for now given that the structure is going to change a bit with the implementation of FHIR conversion, so I think any changes from classes to functions should happen in future iterations. At this point, this PR is basically for historical/documentation purposes to show where we started at.

PR #21, opened 2022-02-03
Josiahsiegel/az viz

JosiahSiegel:
This PR adds an example GitHub actions workflow that reads from the Azure resource group and generates a diagram, validating connectivity between GitHub and Azure. Credentials are stored as a repo secret.

A local action is also added to provide common variables, such as environment name or version, when needed.

### DIFF
--- a/.github/actions/build-vars/action.yml
+++ b/.github/actions/build-vars/action.yml
@@ -1,0 +1,51 @@
+# action.yml
+name: 'Build Variables'
+description: 'Variables shared by build workflows'
+outputs:
+  env_name:
+    value: ${{ steps.build_result.outputs.env_name }}
+  version:
+    value: ${{ steps.version_result.outputs.version }}
+
+runs:
+  using: "composite"
+  steps:
+
+    - name: Set Build Environment - DEV
+      id: build_dev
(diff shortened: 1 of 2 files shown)


.github/actions/build-vars/action.yml
```diff
@@ -0,0 +1,51 @@
...
+  env_name:
+    value: ${{ steps.build_result.outputs.env_name }}
+  version:
+    value: ${{ steps.version_result.outputs.version }}
+
+runs:
+  using: "composite"
+  steps:
+
+    - name: Set Build Environment - DEV
```
cwinters-usds, line 14: FWIW I think these will change based on the team's discussion of how they want to do development. That is, there may only be `DEV` and `TEST` for now. (And maybe not even `DEV`?)

cwinters-usds (approved):
LGTM, with the caveat that dev/test/etc. will probably change. But this sets up the framework to represent that change, which is :100:

NatHillardUSDS (approved):
LGTM - thanks for adding! The azure visualization in particular is very cool. Agreed that we might want to eventually remove other environments but seems fine for now

PR #22, opened 2022-02-04
Josiahsiegel/data factory proto

JosiahSiegel:
This PR brings in the **data factory** and all the basic connections that go with it.
If the data factory becomes part of the final setup, then we can also bring additional data factory components, but this will get us started.

TF has also been cleaned up a bit and optimized where possible.

### DIFF
--- a/operations/app/terraform/modules/log_analytics_workspace/main.tf
+++ b/operations/app/terraform/modules/log_analytics_workspace/main.tf
@@ -1,143 +1,44 @@
 resource "azurerm_log_analytics_workspace" "pdi" {
-  name                = "pitest-law"
+  name                = "${var.resource_prefix}-law"
   location            = var.location
   resource_group_name = var.resource_group_name
   sku                 = "PerGB2018"
   retention_in_days   = 30
 }
 
-locals {
-  default = {
-    "function_app" = {
-      id   = "${var.function_app_id}"
-      name = "function_app"
-    },
-    "function_infrastructure_app" = {
-      id   = "${var.function_infrastructure_app_id}"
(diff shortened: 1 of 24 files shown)


operations/app/terraform/modules/log_analytics_workspace/main.tf
```diff
@@ -1,143 +1,44 @@
...
-
-resource "azurerm_monitor_diagnostic_setting" "vnet_diag" {
-  name                       = "vnet_diag"
-  target_resource_id         = var.cdc_managed_vnet_id
-  log_analytics_workspace_id = azurerm_log_analytics_workspace.pdi.id
-
-  log {
-    category = "VMProtectionAlerts"
-    enabled  = true
+  dynamic "log" {
```
cwinters-usds, line 15: TIL :raised_hands:

operations/app/terraform/vars/test/variables.tf
```diff
@@ -81,5 +81,6 @@ variable "use_cdc_managed_vnet" {
 }
 
 variable "terraform_caller_ip_address" {
-  default = "162.224.209.174"
+  type    = list(string)
+  default = ["162.224.209.174/32","73.173.186.141/32","24.163.118.70/32"]
```
cwinters-usds, line 85: are these additional addresses from GitHub or...?

JosiahSiegel: 73 is from @NatHillardUSDS and 24 is myself

cwinters-usds (approved):
Thank you -- this is a lot of wiring!

PR #24, opened 2022-02-07
Adding meta-proposal to use Architectural Decision Records

NatHillardUSDS:
A meta-proposal to use Architectural Decision Records to record decisions on our project﻿


JSkeets (approved):
This is great!

PR #27, opened 2022-02-07
Update DecryptFunction test to use test key

bmaden-usds:
`DecryptFunction` test updates:

* Remove dependence on access to keys stored in key vault
* Add a test for input that fails to decrypt

With this, tests can be run locally with `python -m pytest` without needing to download the pgp key from the key vault.

Fixes https://app.clickup.com/t/1zbqup9


src/FunctionApps/PITest_FunctionApp/DecryptFunction/settings.py
```diff
@@ -8,12 +8,7 @@ class DecryptSettings:
     When run remotely, settings are passed by environment variables in the function app, 
     which in turn reference KeyVault variables as mentioned here: https://docs.microsoft.com/en-us/azure/app-service/app-service-key-vault-references
 
-    When run locally, these consult a local file, local.settings.json.
-    This can be copied from the Azure portal via:
-        ```func azure functionapp fetch-app-settings pitest-functionapp --output-file local.settings.json
-            func settings decrypt
-        ```
-        to get the values direct from the portal, then edited to set the variables.
+    When tests are run, they consult the local file `test.settings.json`.
```
NatHillardUSDS, line 11: Maybe a short note here that this represents a different key than that stored in the keyvault ?

NatHillardUSDS (approved):
Awesome, thanks again for this awesome catch!

PR #28, opened 2022-02-08
Update Getting Started Doc to Include Java/Kotlin

bryan-skylight:
# Description

This pull request makes a few changes to the `getting_started.md` file:

* Adds more details or further clarifies points in various sections.  
* Rearranges some of the sections to provide a better flow, starting with a more general review of using VS Code and then diving into the specifics of working with Python and Java.  
* Removes references to 'golden path' as this is biased language in my opinion. The language implies that using a Mac is better than using Windows for development on this project, but the tech stack is platform agnostic.  
* Adds details on developing in Java and Kotlin, including setting up environment variables, defining dependencies, and installing necessary software, such as Maven.  

# Screenshots
Not necessary, as the doc can be view in the PR

# Feedback Desired

* Do others agree that the document flows better like this? If not, we can revert to the previous layout and I can append my changes towards the bottom.  
* Do the instructions make sense, and do you feel confident that you'd be able to set up your local environment using these instructions? If not, which parts are unclear?  
* I know that we still need documentation on creating tests for Azure Functions in Java. That update will be coming shortly after a few kinks are ironed out and the process is better understood.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -41,34 +41,180 @@ The below instructions cover how to setup a development environment for local de
 
 ### Hardware
 
-Until we have properly containerized our apps, we will need to rely on informal consensus around hardware. As things stand, the "golden path" involves using VSCode on a Mac. Notably, M1 macs are unable to run the `azure-functions-core-tools` package directly (and even have difficulty running it through Rosetta), so please seek out an alternative machine for now!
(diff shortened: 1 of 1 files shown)


bamader (approved):
This is great! I haven't set up my local dev environment yet so I'll be a good proof of concept test for someone coming in to do this blind. Assuming this all works, do we feel that this PR also satisfied this ticket: https://app.clickup.com/t/27bb01u ?

PR #31, opened 2022-02-09
Add workflow to execute tests and code checks

bmaden-usds:
This PR adds a workflow that will run when pull requests are created.  It will run:

* Tests on python code using `pytest`
* Tests and other checks on kotlin code with `mvn verify`
* Formatting and style checks on python code
* Formatting checks on terraform

For this workflow to complete successfully, this depends on merging the following branches:

* `task-25qr6zp` (add DSTP function app)
* `bmaden-usds/format` (apply python and terraform formatting)

An example completed run: https://github.com/CDCgov/prime-public-health-data-infrastructure/actions/runs/1820655925

Fixes https://app.clickup.com/t/28ymg61

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -4,9 +4,4 @@ exclude =
     .pytest_cache,
     __pycache__,
     docs
-ignore =
-    # line too long: ignore for now
-    E501,
-    # binary op on new line: combo of pathlib and black causes this
-    W503
 max-line-length = 88
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,12 @@
+[flake8]
+exclude =
+    .git,
+    .pytest_cache,
+    __pycache__,
+    docs
+ignore =
+    # line too long: ignore for now
+    E501,
+    # binary op on new line: combo of pathlib and black causes this
+    W503
+max-line-length = 88
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -9,6 +9,7 @@ env:
   TEST_RUNNER_JAVA_VERSION: 8
   TEST_RUNNER_NODE_VERSION: 14
   TEST_RUNNER_PYTHON_VERSION: 3.9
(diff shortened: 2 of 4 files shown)


.flake8
```diff
@@ -0,0 +1,12 @@
...
+    .git,
+    .pytest_cache,
+    __pycache__,
+    docs
+ignore =
+    # line too long: ignore for now
+    E501,
+    # binary op on new line: combo of pathlib and black causes this
+    W503
+max-line-length = 88
```
cwinters-usds, line 12: Why include a max-line-length if we're ignoring? (Also curious about where 88 came from, there are so many ideas around what this should be!)

.github/workflows/test.yaml
```diff
@@ -0,0 +1,109 @@
...
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./operations
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: hashicorp/setup-terraform@v1
+        with:
+          terraform_version: 1.1.5
```
cwinters-usds, line 107: Maybe it's my mind and foolish consistencies, but can we make this an `env.FORMATTER_TERRAFORM_VERSION` (or whatever) like the other versions?

.flake8
bmaden-usds: This line length matches black's default line length, so I just stuck with that.  The reason I added the ignore for now is that there are some comments that aren't easy to break up into shorter lines (some URLs, if I recall).  Black will mostly output code that adheres to the rules, but there were some exceptions that need to be fixed manually.

There aren't too many of these, so that's a good call-out.  I can swap these over to use `# noqa: E501` and remove this ignore.

NatHillardUSDS (approved):
LGTM! I like the use of the test matrix and overall it's clear and well put-together!

PR #32, opened 2022-02-10
General profiling notebook

bamader:
Since I'll be out tomorrow, I wanted to at least check in this notebook so that Vincent could start applying it to actual data. It doesn't do much (I didn't get to computing stats over all the messages, but you can just use the single profile segment in message function and for-loop over the collection of messages), but it answers some of those general field questions we mentioned in standup. To run it just change the location of the virginia "PRE" data file from Nat's upload (also make sure you've pip installed the package https://python-hl7.readthedocs.io/en/latest/).

### DIFF
--- a/src/notebooks/ELR Profiling Template.ipynb
+++ b/src/notebooks/ELR Profiling Template.ipynb
@@ -1,0 +1,232 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# ELR Data Profiling Template"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This notebook is a template for how we might use some basic python data analysis (i.e. pandas) to profile incoming ELR data. The code is likely adaptable to ECR data as well, if it fits into a pipe-delimited HL7 v2.XX format. This notebook will attempt to answer extremely basic questions involving the cleanliness and formatting of the data, such as proportion of missing fields per field, proportion of missing required fields, proportion of records with demographic info, etc."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 79,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "'''\n",
+    "Note on use: the HL7.Container object parsed by the hl7 package\n",
+    "is a subclass of python list, which gives us easy access for indices:\n",
+    "  Message\n",
+    "  Segment\n",
+    "  Field\n",
+    "  Repetition\n",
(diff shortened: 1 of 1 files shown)


src/notebooks/ELR Profiling Template.ipynb
```diff
@@ -0,0 +1,280 @@
...
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Location of file to open\n",
+    "import os\n",
+    "\n",
+    "HOME_DIR = os.path.expanduser('~')\n",
+    "DATA_DIR = os.path.join(HOME_DIR, 'inventory', 'SampleData', 'PRE') # THIS IS THE ONLY HARD CODE PATH YOU NEED TO CHANGE\n",
+    "IN_FILE = os.path.join(DATA_DIR,'PreTranslation_to_State_VA.hl7')"
```
VincentLaUSDS: This is minor but I just updated the hardcoding of the directory.

src/notebooks/.ipynb_checkpoints/ELR Profiling Template-checkpoint.ipynb
```diff
@@ -0,0 +1,280 @@
...
+    "            strp_line = line.strip()\n",
+    "            msg += '\\r' + strp_line\n",
+    "\n",
+    "# Verify correctness of data loading: I count ten messages in this \n",
+    "# representative set; all messages MUST begin with the message\n",
+    "# header signifier 'MSH'\n",
+    "expected_msgs = 10\n",
+    "assert len(msgs) == expected_msgs\n",
+    "starts_with_msg_header = [x.split('|')[0].strip() == 'MSH' for x in msgs]\n",
+    "assert all(starts_with_msg_header)\n",
```
VincentLaUSDS, line 80: I see, so this assertion is hardcoded to 10 based on that specific file right? So if we were to run this in production on real Virginia data we should take this out?

VincentLaUSDS (approved):

src/notebooks/.ipynb_checkpoints/ELR Profiling Template-checkpoint.ipynb
bamader: Correct! In prod, we could a) change this to an assertion as to the length of whatever snapshot we're looking at, or b) eliminate this and just stick with the sanitation check that all messages start with `MSH`

VincentLaUSDS: Great!

PR #33, opened 2022-02-11
Apply python and terraform formatting

bmaden-usds:
Format all existing python and terraform code.

* python is formatted with `black` in it's default configuration
* terraform is formatted with `terraform fmt`

Later, this will allow the code formatting and style checks that are in https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/31 to pass.

This is a big diff, but there are very few real changes, these can be seen in: https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/33/commits/b529959c37b0e791fef030a3ca01a6b5c08a0b55


NatHillardUSDS (approved):
Nice, looks good to me, thanks for running this! We may want to add pre-commit or similar to run this automatically but we'd want to discuss this first; for now though this is great to ensure tests don't fail for old files

PR #34, opened 2022-02-11
Update DecryptFunction to use HTTP trigger

NatHillardUSDS:
Converts this function to an HTTP trigger, which is a bit easier to work with  and fire on demand. Also adds additional tests for expected inputs and outputs


src/FunctionApps/PITest_FunctionApp/tests/test_decrypt_function.py
```diff
@@ -77,3 +78,74 @@ def test_decrypt_message_failure_wrong_receiver(local_settings):
...
+
+
+def test_trigger_success(local_settings):
+    """Test function trigger conditions
+
+    Args:
+        local_settings ([type]): passed automatically via above fixture
+    """
+    test_file_path = (
+        files("DecryptFunction").parent / "tests" / "assets" / "encrypted.txt"
```
bryan-skylight, line 90: Do you need to change this to `Path("DecryptFunction").parent ...` like you did on line 27?

src/FunctionApps/PITest_FunctionApp/tests/test_decrypt_function.py
```diff
@@ -77,3 +78,74 @@ def test_decrypt_message_failure_wrong_receiver(local_settings):
...
+        local_settings ([type]): passed automatically via above fixture
+    """
+    req_success = func.HttpRequest(
+        method="POST",
+        body=b"bad data",
+        headers={"Content-Type": "application/octet-stream"},
+        url="/",
+    )
+    resp = dcf.main(req_success, {})
+    assert resp.status_code == 500
```
bryan-skylight, line 151: Can we test that it's failing for the right reason? There are a couple ways a status code of 500 is returned, so I think it would be better to test that the status code is 500 **and** the response is "server settings missing".

src/FunctionApps/PITest_FunctionApp/tests/test_decrypt_function.py
```diff
@@ -77,3 +78,74 @@ def test_decrypt_message_failure_wrong_receiver(local_settings):
...
+        local_settings ([type]): passed automatically via above fixture
+    """
+    req_success = func.HttpRequest(
+        method="POST",
+        body=b"bad data",
+        headers={"Content-Type": "application/octet-stream"},
+        url="/",
+    )
+    resp = dcf.main(req_success, local_settings)
+    assert resp.status_code == 500
```
bryan-skylight, line 135: Can we test that it's failing for the right reason? There are a couple ways a status code of 500 is returned, so I think it would be better to test that the status code is 500 **and** the response is "decryption failed".

graham-skylight (approved):
looks good!

src/FunctionApps/PITest_FunctionApp/tests/test_decrypt_function.py
NatHillardUSDS: Good callout - thanks! Graham's recent change also used this [here](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/42/files). I do think we'll want to fix this more holistically in the future / using package structure, but if this works for now in VSCode, CLI, and when running we should use it instead

src/FunctionApps/PITest_FunctionApp/tests/test_decrypt_function.py
NatHillardUSDS: Very good point! I've fixed this.

src/FunctionApps/PITest_FunctionApp/tests/conftest.py
```diff
@@ -0,0 +1,12 @@
+import json
```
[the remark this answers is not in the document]
NatHillardUSDS: Good call, removing!

PR #35, opened 2022-02-11
RFC: Add cached geocoding example & docs

graham-skylight:
# Example Function App w/ Caching

This PR adds an example geocoding function app (using the Census endpoint) while caching the lookups in redis for a week.

## Questions

* Is a function app always going to be the best way to do this?
* Can we think of a way to better integrate this with the notebooks?

## Azure Setup Screenshots

The README has more detailed instructions to set up an Azure environment - these screenshots are the not-super-obvious parts involving the config.

Grabbing connection values from Redis

redis-settings-screenshot

Adding them to the function app

function-config-screenshot


NatHillardUSDS:
Following on to your arch question: I'm curious about serverless alternatives to Azure Cache for Redis; this [bills by the hour](https://azure.microsoft.com/en-us/pricing/details/cache/), rather than on-demand. We're obviously up against two paradigms - persisted + stateless - but I am curious if there might be a way to access a cache on demand? 

I know global objects can potentially persist between function invocations; there are also [durable entities](https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-entities?tabs=python) which do this. This might be an avenue at least worth comparing, so as to avoid an always-on server?

graham-skylight:
Durable entities sound like they could be worth a shot - I also ran across Cosmos (https://azure.microsoft.com/en-us/pricing/details/cosmos-db/) which has a 'serverless' option where they price per request rather than instance hours

graham-skylight:
I looked into Cosmos and durable entities a bit:

## Cosmos DB

Cosmos has a serverless option where we'd pay by the 'request unit' - something like $0.30 / million smallish requests. They've got Cassandra and MongoDB drivers and a TTL option to expire keys.

## Durable Entities

It looks like we could hack something together, but I ran into a few issues with needing a storage account. Beyond that, it's not obvious how you can expire keys, so it looks like something we'd have to do manually. Also, I'd suspect that it's probably fine, but I haven't been able to find a ton about performance (we'd be using it as a cache layer after all).


All of that said, do we have enough information to work towards an actual implementation?

bmaden-usds (commented):
To run the tests, you should be able to rebase or merge main and then add `CachedGeocode` to the list of functions to test in `.github/workflows/test.yaml:20`.

This was just added by a PR I merged, so I'd also like to make sure it's working : )

graham-skylight:
FYI - the github actions failure is trying to test the DSTP function app from #18

bmaden-usds:
> FYI - the github actions failure is trying to test the DSTP function app from #18

Do we think that will get merged soon?  If so, I'd say just let this fail until it gets merged.

However.... green check marks are nice.  We could just disable the java/kotlin checks for now.  An easy way to do that would be:

```
diff --git a/.github/workflows/test.yaml b/.github/workflows/test.yaml
index d03a693..6f001b5 100644
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -40,6 +40,7 @@ jobs:
           python -m pytest
 
   unit-test-java:
+    if: ${{false}}  # disable until at least 1 kotlin function app is available to test
     runs-on: ubuntu-latest
     strategy:
       matrix:
```

graham-skylight:
Agreed - I'm fine letting it fail for now (just wanted to let other reviewers know why there's a big red x). I think the big question remaining is using redis vs cosmos and how far we want to go for the initial spike

NatHillardUSDS:
Yeah my thoughts here are that we should merge this current ticket and create a separate spike on the tradeoffs of the other potential options; meanwhile once we decide we can have another to handle adding this to terraform?

PR #36, opened 2022-02-17
Spike: Use Cosmos DB as a serverless cache

graham-skylight:
# Overview

This modifies the redis cache from #35 to use the serverless version of Cosmos DB as a caching layer.

## VS Redis

These endpoints typically return in 300-500ms for a cache hit (vs Redis' ~100ms), but that's almost certainly due to the serverless option.

However, this probably comes into play if the geocoding service we're using is relatively fast/cheap - if it regularly beats 300ms for a geocoding lookup, then caching in Cosmos DB would actually slow things down. At that point we'd want to look into either booting a regular instance or using Redis.


JSkeets:
I honestly have little preference in Redis vs Cosmos. I feel we make a decision stick with it, and move on. If we need performance improvements if the project grows we can address them then. It seems that Cosmos will allow us to get off the ground quicker, which seems like a win to me.

bamader (approved):
Echoing what Jesse said, +1 for Cosmos. Having proofed the geocoding code and done some runs with Vincent, it like never beats the 300ms mark so caching should generate some performance savings.

graham-skylight:
Thanks all - going to merge this and switch the ADR to pick Cosmos

PR #37, opened 2022-02-18
Add draft ADR for caching

graham-skylight:
Open a discussion about the outcome of the caching spikes (#35 , #36)

### DIFF
--- a/docs/decisions/0003-caching-layer.md
+++ b/docs/decisions/0003-caching-layer.md
@@ -1,0 +1,59 @@
+# 3. Select a caching layer for repetitive lookups
+
+Date: 2022-02-18
+
+## Status
+
+Draft
+
+## Context and Problem Statement
+
+Several data augmentations may end up being somewhat to very repetitive, and a caching layer may help with
+
+- Speeding up repeated actions
+- Reducing the cost of external services
+
+## Decision Drivers
+
+- **Simplicity** - We want to be able to reason about the architecture without cluttering things up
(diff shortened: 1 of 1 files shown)


bamader (approved):
Personally, I like all the options except 2. I'm not concerned with the performance hits of Cosmos, but am very concerned about request pricing for development purposes. I like Azure Cache in general, but managing extra infra goes against my personal belief that we should make the minimal infrastructure choices we can to get things managed. Finally, for Python's LRU cache, it's unclear to me if batch caching would actually be a big problem, but that would likely depend on our choice of implementation. Great summary putting this together, though, and I particularly love the articulation of the decision drivers!

graham-skylight:
Yep! I think I personally like Redis & Cosmos pretty much equally (infra vs :shrug: pricing basically). I think the LRU cache would be nice, but only if we'd be able to guarantee a batch run - if that's what we end up doing then things get way easier.

bryan-skylight (approved):
Overall, this looks great to me. Clear, concise, and lays out great options for consideration. My only question is whether we should add more details to the Decision Outcome section, possibly noting who weighed in on the decision, or adding links to conversations? I don't know if there's value in this given that we could simply use the PR review approvals as a proxy for who weighed in, and I don't want to create work if there's no value. I just wanted to raise the idea to see what others thought.

graham-skylight:
I think I'm generally in favor of looking up the PR if we need that level of granularity, but also I tend to think of ADRs as more about the decision rather than the deciders if that makes sense.

PR #38, opened 2022-02-18
Add pre-commit config to run checks locally

bmaden-usds:
For thought and discussion.  There don't seem to be any real downsides for using local checks using `pre-commit` since it only adds optional configuration changes to the local dev environment.

This PR makes two changes:

1. Add a [`pre-commit`](https://pre-commit.com/) configuration to run the code checks locally that are already run on PRs (https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/31).  This runs `black` and `flake8` on python code and `terraform fmt` on terraform code.  `terraform` may not be installed locally by all developers and in that case, this terraform check will pass allowing this config to be used even if terraform isn't installed.  This check was done instead the available terraform pre-commit repos because all of them I checked fail when terraform is not available.
2. Run test workflow on push to `main`, this will trigger it to run after PR merges and in any case where code is pushed directly to `main`

Some info about using pre-commit has been added to [docs/getting_started.md](https://github.com/CDCgov/prime-public-health-data-infrastructure/blob/bmaden-usds/pre-commit/docs/getting_started.md#pre-commit-optional) to describe how to set up and use pre-commit.

If pre-commit is not installed or if you do not install the hooks, then these changes will not affect you.


NatHillardUSDS (approved):
This is awesome, thanks again Brett for a great change! One thing that has come up in discussion: we probably don't want the python code in our notebooks to be formatted via Black; is it possible to add `src/notebooks` to an exception list for both `precommit` and the github action?

bmaden-usds:
> is it possible to add `src/notebooks` to an exception list for both `precommit` and the github action?

Yes, definitely!  I pushed another commit into this PR that makes that change.  It adds a `black` configuration that applies to both the github workflow and running locally.  I went ahead and updated the workflow so it doesn't install the jupyter extras for black since it is no longer necessary.

PR #41, opened 2022-02-23
Update Java Test Runner to Java 11

bryan-skylight:
# Definition
Updates the Java version from 8 to 11 in the Github actions so that `java.net.http` is supported.

# Screenshots
N/A

# Feedback
Does this automatically use Java 11 on the backend? There's nothing else we need to do?


NatHillardUSDS (approved):
LGTM!

PR #42, opened 2022-02-24
Use pathlib instead of importlib for test fixtures

graham-skylight:
Should now work in vscode and cli - :crossed_fingers: for github actions

### DIFF
--- a/src/FunctionApps/PITest_FunctionApp/tests/test_decrypt_function.py
+++ b/src/FunctionApps/PITest_FunctionApp/tests/test_decrypt_function.py
@@ -1,6 +1,5 @@
 import json
-from importlib.resources import files
-
+import pathlib
 
 import azure.functions as func
 import pgpy.errors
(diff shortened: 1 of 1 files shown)


bryan-skylight:
This looks pretty straightforward. Are there steps to replicate the issue that this PR solves so that we can test the before and after? Or is the fact that all of the checks ran sufficient proof that it works?

graham-skylight:
The issue was that using `pytest` from a virtualenv (as opposed to `python -m pytest` which worked fine) wouldn't work with the importlib imports.

The tl;dr was that certain environments search the current directory for imports and some don't. As long as this works however people would normally run tests, I think we're good.

NatHillardUSDS (approved):
Hm, technically `importlib`, at least `importlib.resources` is meant to handle this case ([ref](https://realpython.com/python-import/#introducing-importlibresources)) by doing package-relative imports, but if this works for local tests and bot tests let's do it!

bamader (approved):
Seems legit!

graham-skylight:
@NatHillardUSDS yep - the issue is relying on `$PWD` being in the import path, which depends on the tool being used. I figured using pathlib and `__file__` instead of messing with the import path was a little easier

PR #43, opened 2022-02-24
Add flatten and explode

bamader:
This PR adds a notebook to handle the transformation of ndjson data into a spark dataframe format for future use in databricks and analysis with parquet. Columns that have embedded structs or array-likes as their values are expanded and exploded, respectively, so that new rows are added to the resulting data frame for each element encountered (this makes the result fully 'flat').


VincentLaUSDS:
What's the best way to test this? Did you create this in Databricks and then just copy and pasted the content to a local .ipynb file?

src/notebooks/flatten_and_explode.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "def explode_arrays(df):\n",
+    "    flat_cols = [field.name for field in df.schema.fields if type(field.dataType) != ArrayType]\n",
+    "    struct_cols = [field.name for field in df.schema.fields if type(field.dataType) == ArrayType]\n",
+    "    \n",
+    "    # Add new rows for each array element, then select and join with columns\n",
+    "    # that are already flat\n",
+    "    exploded_df = df.withColumn('vals', explode_outer(arrays_zip(*struct_cols))) \\\n",
+    "           .select(*flat_cols,'vals.*') \\\n",
+    "           .fillna('', subset=struct_cols)\n",
+    "    return exploded_df"
```
bryan-skylight, line 216: nit: We can make this section of code a little cleaner using the fact that Python considers everything inside of parentheses to be a single line.

```
exploded_df = (
    df.withColumn('vals', explode_outer(arrays_zip(*struct_cols)))
    .select(*flat_cols, 'vals.*')
    .fillna('', subset=struct_cols)
)
```
I think this makes the code cleaner, easier to read, and less prone to errors resulting from for getting the `\` on a line. But there's nothing wrong with how you've implemented it, so go with what you like.

src/notebooks/flatten_and_explode.ipynb
```diff
@@ -0,0 +1,524 @@
...
+   ],
+   "source": [
+    "'''\n",
+    "A wrapper function to perform full flattening of an input spark data frame.\n",
+    "Struct columns are separated out into multiple columns, and then the intermediate\n",
+    "expanded data frame is exploded so that each new element encountered generates a \n",
+    "new row.\n",
+    "'''\n",
+    "def flatten_df(dfflat):\n",
+    "    while len([field.name for field in dfflat.schema.fields if type(field.dataType) == StructType or type(field.dataType) == ArrayType ]) !=0 :\n",
```
bryan-skylight, line 273: You can make this a little cleaner with the following:

```
while len([field.name for field in dfflat.schema.fields if type(field.dataType) in [StructType, ArrayType]]) != 0:
```

Additionally, while not really an issue here, I think it's best to get into the habit of being explicit about what conditions you're testing for. Here, you've said that as long as the length does not equal 0 then lines 274 and 275 should run, which has an implicit understanding that `len` can only be positive. Implementing these implicit assumptions can lead to unforeseen outcomes and I'd recommend trying to avoid them where possible. To that end, I'd rewrite this to be:

```
while len([field.name for field in dfflat.schema.fields if type(field.dataType) in [StructType, ArrayType]]) > 0:
```

src/notebooks/flatten_and_explode.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "  CREATE TABLE [table_name] (\n",
+    "    [field1] nvarchar(50),\n",
+    "    [field2] nvarchar(50),\n",
+    "    ...\n",
+    "  )\n",
+    "  \n",
+    "For type safety reasons, all fields in the created DDL are saved as nvarchar(50).\n",
+    "'''\n",
+    "def generate_ddl(df, tblname):\n",
+    "    createtbl = 'CREATE TABLE [' + tblname +  '] ( \\n'\n",
```
bryan-skylight, line 339: I think f-strings would provide a bit more readability.

```
createtbl = f"CREATE TABLE [{tblname}] ("
```

I also recommend leaving the `\n` out of the string as we can concatenate everything at the end, again to improve readability.

src/notebooks/flatten_and_explode.ipynb
```diff
@@ -0,0 +1,524 @@
...
+    "    num_columns = len(dfflat.columns)\n",
+    "    for i, y in enumerate(dfflat.columns):\n",
+    "        if i == num_columns-1:\n",
+    "            column_name = '\\t['+ y + '] nvarchar(50)); \\n\\n'\n",
+    "        else:\n",
+    "            column_name = '\\t[' + y + '] nvarchar(50), \\n'\n",
+    "        createtbl += column_name\n",
```
bryan-skylight, line 346: I would replace lines 340 to 346 with the following:

```
fields = ", \n".join([f"\t[{column}] nvharchar(50)" for column in dfflat.columns])
createtbl = "\n".join([createtbl, fields, ";"])
```

This approach appends the semi-colon on a newline by itself, but SQL doesn't care about that and it improves the readability of the code.

bryan-skylight (changes requested):

DanPaseltiner:
CSV generation for the VA pilot was done using a different approach. As we begin to transform JSON FHIR bundles into parquet this work may be revisited.

PR #44, opened 2022-02-25
ADR for spike: where should the FHIR server live in our pipeline?

NatHillardUSDS:
Adding ADR for #1zx036x[review] : where should the FHIR server live in our pipeline?


docs/decisions/0004-fhir-server-architecture.md
```diff
@@ -0,0 +1,146 @@
...
+              Curation --> Standardization
+   ```
+
+   1. Pros:
+      1. **Simplicity**
+         1. The FHIR server links our per-resource requests, at first "soft linking" via references and then later hard-linking by using "if-match" [criteria](https://www.hl7.org/fhir/http.html#concurrency)
+         2. It provides out of the box storage and a managed database
+         3. It provides ready-made functions that we can use without needing to re-implement or re-host in the short-term:
+            1. `$convert-data` - HL7 -> FHIR
+            2. `$validate` - apple FHIR profile to data
```
graham-skylight, line 95:
```suggestion
            2. `$validate` - apply FHIR profile to data
```

graham-skylight (approved):
Nice work with the mermaid diagrams, LGTM

docs/decisions/0004-fhir-server-architecture.md
```diff
@@ -0,0 +1,146 @@
...
+         1. The FHIR server links our per-resource requests, at first "soft linking" via references and then later hard-linking by using "if-match" [criteria](https://www.hl7.org/fhir/http.html#concurrency)
+         2. It provides out of the box storage and a managed database
+         3. It provides ready-made functions that we can use without needing to re-implement or re-host in the short-term:
+            1. `$convert-data` - HL7 -> FHIR
+            2. `$validate` - apply FHIR profile to data
+            3. `$export` - export data
+            4. `$match` - annotate linked records
+            5. Through [FHIR Proxy](https://github.com/microsoft/fhir-proxy), Bulk Insertion and pre/post-processeing. We would need to add this on separately
+      2. **Consistency** - this lets us work with FHIR data right from the start of our pipeline and then treat it as flat files going forward
+      3. **Composability** - having this functionality encapsulated in specific endpoints could let us pel these off if needbe
```
bryan-skylight, line 100: typos: "...endpoints could let us **peel** these off if **need be**

bryan-skylight:
Thanks for putting all of this together, Nat. It looks great. Seeing it laid out like this, though, I have a couple follow up questions before accepting the PR.

1. Are we thinking about this in the right way? 

The proposed solution implies that the FHIR server sits in a single spot within the pipeline, but I'm not sure that's necessarily true. For example, the first thing we do with the decrypted data is bounce it against the FHIR Server to convert it to a FHIR bundle. And as your diagram lays out, this is true regardless of where the FHIR Server sits, with the exception of the first scenario in which we'd use Microsoft's FHIR Converter CLI. This leads me to my next question.

2. Does it make more sense to ask "which services should be performed by the FHIR server" instead of "where should the FHIR Server sit in our pipeline"? 

This might seem like a distinction without a difference, but the latter question implies the data runs through the FHIR Server once, whereas the former question implies that it can be called any number of times. Looking at it from a services perspective, we can ask more detailed questions like:

    - How should we convert HL7 v2 and v3 data to FHIR? Should use the FHIR Server's `convert-data` endpoint, or the CLI?
    - How should we perform record linkage on the data? Should we do it fully by hand? Lean entirely on the FHIR Server's `if-match` capabilities? Or use a hybrid?
    - How should we convert the data into the proper flat-file format? Should we use Python's native ability to write to Parquet files? Or should we use the FHIR Server's `export` endpoint? 
    - How should we represent relationships between Resources like, say, Patient and Physician? Should we lean Azure's managed FHIR Server to map those connections? Or should we do it manually?

I think we know the answers to all of these questions, but I think those questions better get at the root of what we're trying to do and solve.

JSkeets (approved):

bamader (approved):

docs/decisions/0004-fhir-server-architecture.md
NatHillardUSDS, line 100:
```suggestion
      3. **Composability** - having this functionality encapsulated in specific endpoints could let us peel these off if need be
```

NatHillardUSDS:
> Thanks for putting all of this together, Nat. It looks great. Seeing it laid out like this, though, I have a couple follow up questions before accepting the PR.
> 
> 1. Are we thinking about this in the right way?
> 
> The proposed solution implies that the FHIR server sits in a single spot within the pipeline, but I'm not sure that's necessarily true. For example, the first thing we do with the decrypted data is bounce it against the FHIR Server to convert it to a FHIR bundle. And as your diagram lays out, this is true regardless of where the FHIR Server sits, with the exception of the first scenario in which we'd use Microsoft's FHIR Converter CLI. This leads me to my next question.
> 
> 2. Does it make more sense to ask "which services should be performed by the FHIR server" instead of "where should the FHIR Server sit in our pipeline"?
> 
> This might seem like a distinction without a difference, but the latter question implies the data runs through the FHIR Server once, whereas the former question implies that it can be called any number of times. Looking at it from a services perspective, we can ask more detailed questions like:
> 
> * How should we convert HL7 v2 and v3 data to FHIR? Should use the FHIR Server's `convert-data` endpoint, or the CLI?
> * How should we perform record linkage on the data? Should we do it fully by hand? Lean entirely on the FHIR Server's `if-match` capabilities? Or use a hybrid?
> * How should we convert the data into the proper flat-file format? Should we use Python modules to write to Parquet files? Or should we use the FHIR Server's `export` endpoint?
> * How should we represent relationships between Resources like, say, Patient and Physician? Should we lean Azure's managed FHIR Server to map those connections? Or should we do it manually?
> 
> I think we know the answers to all of these questions, but I think those questions better get at the root of what we're trying to do and solve.

This is a really excellent line of questioning; I would wonder in this case if we can for now use the FHIR Server, knowing there may be benefit to moving off of this in the future, but that to do so we'd need to host separate services? I think if we separate out our concerns enough at the outset / draw the right microservices boundaries, we can change out the implementation in the future. 

As an example, while now we're calling into the FHIR server for FHIR conversion, this is a shortcut for hosting the FHIR-Converter source itself and calling that. I would say the one major benefit of the server at the moment is that instead of standing up 5 separate services (in addition to the function apps themselves, we need to somehow host/run the "Validation" jar for instance), we are able to use the server out of the box. I can make this more explicit in the description though

PR #47, opened 2022-03-01
[cu-2fr7t8x] Auto-detect function apps to test

bmaden-usds:
Fixes https://app.clickup.com/t/2fr7t8x

### Description

Scan the directory `src/FunctionApps` and look for directories containing a function app that needs to have tests run on it.

This will remove the need to update the workflow when a function app is added or removed.

#### Determining language

This is done by looking into each top-level directory under `FunctionApps` for a file.  If a `requirements.txt` file exists, the directory is determined to have a python function app in it.  Java/Kotlin apps are contain a `pom.xml` file (the workflow currently only supports maven).  Lists of the found function apps are used to set the matrix of python and java apps to tests.

#### Workflow modification

1. Add a new job that generates lists of function-app directory names and sets them as outputs
2. Make the unit test matrix jobs dependent on the new job
3. Use the outputs from the new job to set the matrix parameter for each unit test job

### Notes

The scan outputs a list of only the directory names.  This makes detecting the function apps a little bit messier, but the benefit is that it is easy to generate nice job names for the workflow output.  For the time being, they will remain the same as they were before this modification, for example: `unit-test-python (FHIRExport)`.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -33,12 +33,13 @@ jobs:
           sudo apt-get install jq
       - id: find-python-apps
         name: Scan for python function apps
+        # use jq to produce json output and filter out the empty item caused by final newline
         run: |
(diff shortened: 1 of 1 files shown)


graham-skylight (approved):
:+1: looks good to me, thanks!

PR #48, opened 2022-03-04
RFC: Transformations between bronze and silver

graham-skylight:
A quick draft of a potential way to transform data between the bronze and silver tiers


src/FunctionApps/SilverTransforms/Transform/__init__.py
```diff
@@ -0,0 +1,66 @@
...
+import pymongo
+
+import azure.functions as func
+
+from Transform.fhir import get_patient_records, write_patient_records
+from Transform.geo import get_smartystreets_client
+from Transform.transforms import transform_record
+
+
+CACHE_EXPIRATION = 60 * 60 * 24 * 7  # seconds * minutes * hours * days = 1 week
```
bamader, line 13: Curious, what's the rationale for setting the cache expiration so high? Is it just under the assumption there will be roughly one week between receiving batches of data?

src/FunctionApps/SilverTransforms/Transform/fhir.py
```diff
@@ -0,0 +1,40 @@
...
+import json
+
+from typing import Iterable
+
+from azure.storage.blob import ContainerClient
+
+
+def get_patient_records(conn_str: str, container_name: str) -> Iterable[dict]:
+    """
+    Given an bucket connection string and a container name,
```
bamader, line 11: Nittiest possible nit: Should be `a`, not `an` 😂

src/FunctionApps/SilverTransforms/Transform/transforms.py
```diff
@@ -0,0 +1,67 @@
...
+    raw = raw.strip()
+    return raw
+
+
+def transform_phone(raw: str) -> str:
+    """Make sure it's 10 digits, remove everything else"""
+    raw = [x for x in raw if x.isnumeric()]
+    raw = "".join(raw)
+    if len(raw) != 10:
+        raw = None
```
bamader, line 21: Hmm can we come up with a more elegant way to handle phone numbers? I don't know if there is one but throwing out all the instances that don't match 10 numbers exactly feels...I don't know, wasteful? Although, if I talk through it, I don't know if there really is an alternative. If it's less than 10 we probably should throw it out (because we can't just make up the missing number), and if it's over 10, how do we choose which 10 to use (could be the first ten maybe)? Idk, I might be rambling, but just flagging in case there are conversation thoughts. If not, happy to resolve this and fade back into the shadows 🐍

src/FunctionApps/SilverTransforms/Transform/transforms.py
```diff
@@ -0,0 +1,67 @@
...
+def transform_record(
+    cache: pymongo.collection.Collection,
+    client: smartystreets_python_sdk.us_street.Client,
+    patient: dict,
+) -> dict:
+
+    # Transform names
+    for i, name in enumerate(patient["name"]):
+        raw_last_name = name["family"]
+        patient["name"][i]["family"] = transform_name(raw_last_name)
```
bamader, line 34: What's the structure of the `patient` dict? Is this supposed to be a dictionary containing _all_ patients (i.e. the first level of field down is a property, then the next field is patient `n` out of `num_patients`)? Based on the variable name I thought it would be a single patient so a little unclear on the structure of this object.

bamader (approved):
Code looks great! Cleanly written and readable (I don't normally point out stylistic things in code writing, but Graham your python style is extremely similar to mine so I found this really nice to read 😂 ). Just a couple pedagogical questions but nothing big or really even blocking.

src/FunctionApps/SilverTransforms/Transform/transforms.py
DanPaseltiner: Definitely happy to think about more sophisticated phone number handling. The challenge is that a US phone number that isn't 10 digits is not unique and therefore not ideal for record linkage. I'd be happy to consider building out functionality to parse different specific formats. This would allow us to identify an extension and choose how we want to handle it.

src/FunctionApps/SilverTransforms/Transform/transforms.py
DanPaseltiner: It is a single patient, but a single patient may have more than one name (name changes). The FHIR patient resource stores first and last name in lists to allow for this. Most of the the time these lists will only have one element. We could just alway pick the most recent name (first element), which may be what we do for linkage, but it seemed easy enough to just standardize all names while interacting with the record.

src/FunctionApps/SilverTransforms/Transform/__init__.py
graham-skylight: Just a random guess here until we know more about what the data looks like (and how we're going to use it) - I know SmartyStreets gets expensive fast, so we may want to bring it up even higher.

src/FunctionApps/SilverTransforms/Transform/fhir.py
graham-skylight: :heavy_check_mark: thanks!

src/FunctionApps/SilverTransforms/Transform/__init__.py
bamader: Totally fair.

src/FunctionApps/SilverTransforms/Transform/transforms.py
bamader: Maybe the best solution is to accept this for now as an easy/strong starting point, and then as we work on refining record linkage, we can just refine phone number parsing under the hood (if it'll be wrapped in a function that'll allow micro changes without affecting larger code). Extension parsing could be a good follow up ticket.

bamader:
All conservations resolved, feel free to merge whenever

JSkeets (approved):
LGTM!

src/FunctionApps/SilverTransforms/Transform/geo.py
```diff
@@ -0,0 +1,68 @@
...
+class GeocodeResult(TypedDict):
+    key: str
+    address: str
+    zipcode: str
+    fips: str
+    lat: float
+    lng: float
+
+
+def cached_geocode(
```
VincentLaUSDS, line 20: Not blocking PR Merge, but after talking to other CDC folks, I had mentioned the notion of "caching" results because it would make things more efficient. For example, with Lab test data, you will often get repeated patients who test several times, so once you've seen them and confirm they are still listing the same address, you don't have to "re-geocode". There's some issues here in terms of compliance in the sense of whether we're allowed to store patient data over a period of time. But in this case you're not really storing patient data, you're just storing the address and geocoded results themselves. I wonder how that shakes out in terms of a compliance issue?

VincentLaUSDS (approved):

src/FunctionApps/SilverTransforms/Transform/geo.py
graham-skylight: definitely agreed - for that matter though, we really probably need to clear smartystreets more generally. Happy to disable caching until we figure it out if that makes sense

PR #50, opened 2022-03-15
Initial draft of transforms library

graham-skylight:
# Initial draft of transforms lib
ticket: https://app.clickup.com/t/2hq4ugg

## Feedback I'm looking for

* Can we think of better ways to organize the functions (pandas interface? list transforms?)
* How's poetry? Do we gain enough to avoid using the standard requirements.txt pattern?
* How hand-wavey is too hand-wavey (we're building docs but not publishing them, same with an actual python package, etc)?


bamader (approved):
Overall this looks good to me. I think it's a perfectly good implementation of the transforms we've talked about as a minimal footprint library. I've already answered the poetry question in a comment, so I'll just deal with function organization and hand wavey.

Re: functions. I think this is perfectly fine actually. While, yes, pandas interfaces would be an ideal, imo with functionality this basic it simply isn't worth actually writing any kind of connection there. Any data scientist will be accustomed to invoking a "pre-processing library" so to speak on data before they shove it into pandas, or whatever other manipulation system they're using. As long as the individual functions themselves are modular, I don't think we need to do anything more than just add standardizations and transforms as we come up with them.

Re: hand-wavey: I'm not perfectly sure what you mean by this question, but if you're asking if there are any other documentation steps we need to take, I would say no. Sphinx gives plenty of results around documentation, and the code itself has all the docstring you'd need for these basic functions, so I don't think we need to introduce anything more.

All in all, then, just a few logic questions outstanding, but nothing in here that I would majorly suggest changing (other than potentially removing poetry).

src/lib/phdi-transforms/docs/conf.py
```diff
@@ -0,0 +1,57 @@
...
+# -- Path setup --------------------------------------------------------------
+
+# If extensions (or modules to document with autodoc) are in another directory,
+# add these directories to sys.path here. If the directory is relative to the
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+#
+import os
+import sys
+
+sys.path.insert(0, os.path.abspath(".."))
```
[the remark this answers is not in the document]
graham-skylight: Yep - we want the next one up - `.` is `docs/`, but `..` would be the parent (which includes the actual code)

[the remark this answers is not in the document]
bamader: DOH totally missed that this was a conf in the docs folder. Makes perfect sense now.

src/lib/phdi-transforms/README.md
```diff
@@ -0,0 +1,15 @@
+# PHDI Transforms Library
+
+## Development
+
+### Installing
+
+This library uses [Poetry](https://python-poetry.org/) to manage dependencies. To get started
+run `poetry install`. After the dependencies are installed, you can either run `poetry shell` to
+activate poetry's virtualenv or `poetry run $YOUR_COMMAND` to run single commands within the virtualenv.
+To run the tests (and black, and flake8), this would be `poetry run make test`.
```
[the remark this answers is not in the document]
graham-skylight: sorry, totally should've elaborated a bit here - I think for our purposes, it's nice that it:

* locks versions for us (alternative would be `pip freeze`, so nbd there)
* handles packaging and publishing to pypi, which we don't want for the prototype, but would be great going forward

graham-skylight:
@bamader thanks for the feedback! I think I've got most of it inline, but to address the hand-wavey stuff - I'm talking about how buttoned up this is going to be for a prototype. Concretely, I'm wondering if the ability to publish docs / packages is enough even though we're not planning on publishing them.

bamader:
@graham-skylight Ahhhh I see. In that case, I would absolutely say yes. We have no plans to publish docs or other packages of any kind, so I think for the prototype especially, there's no reason to invest heavily in them. Based on some of the stuff we talked about in standup about what kinds of analytics shops epis are using, then depending on what research comes back with we can always add more robust doc stuff later. Imo though the _capability_ to publish is itself plenty sufficient.

src/lib/phdi-transforms/phdi_transforms/basic.py
```diff
@@ -0,0 +1,26 @@
...
+
+    raw = [x for x in raw if not x.isnumeric()]
+    raw = "".join(raw)
+    raw = raw.upper()
+    raw = raw.strip()
+    return raw
+
+
+def transform_phone(raw: str) -> str:
+    """Make sure it's 10 digits, remove everything else
```
bryan-skylight, line 16: Have we considered what to do in the face of numbers that include a country code? I presume that this might not even be an issue since all of our data is assumed to come from the US, but even so we should state that somewhere.

src/lib/phdi-transforms/phdi_transforms/geo.py
```diff
@@ -0,0 +1,59 @@
...
+
+class GeocodeResult(TypedDict):
+    """
+    A class representing a successful geocoding response
+    """
+
+    key: str
+    address: str
+    zipcode: str
+    fips: str
```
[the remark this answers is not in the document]
bryan-skylight: I agree that we should consider adding in the county name so as to ensure that analysts and epis have whatever data they are used to working with. In a similar vein, is it worth considering returning the `precision` field from the results? Could be worthwhile for analysts if they have a baseline granularity that they're willing to deal with, but that's purely speculation and I don't know the level of detail returned by the API for that field.

src/lib/phdi-transforms/phdi_transforms/geo.py
```diff
@@ -0,0 +1,59 @@
...
+    client.send_lookup(lookup)
+
+    if lookup.result:
+        res = lookup.result[0]
+
+        addr = [res.delivery_line_1]
+        if res.delivery_line_2:
+            addr.append(res.delivery_line_2)
+
+        return {
```
bryan-skylight, line 36: `city` and `state` are included in the return, but are not defined in the `GeocodeResult` class, and additionally `key` is defined in the class but not included in the output. Does this matter?

src/lib/phdi-transforms/phdi_transforms/basic.py
```diff
@@ -0,0 +1,26 @@
+def transform_name(raw: str) -> str:
+    """trim spaces, capitalize, etc
```
bryan-skylight, line 2: nit: We're converting to uppercase, not capitalizing.

bryan-skylight (approved):
Looks good to me. I have a couple of non-blocking questions, which I've added throughout, but other than that this looks great!

src/lib/phdi-transforms/phdi_transforms/basic.py
bamader: That's actually a really good call out. A lot of automated web pages that intake data (just as an example) automatically concatenate either `1` or `+1` to the front of phone numbers. Numbers that begin with a leading `+` are easy to identify a country code on, but we should have logic in place to handle an 11 digit number that starts with a leading 1. (Perhaps, if the original number had a delimiter between the first one and the next number, and the first one was alone, etc.)

DanPaseltiner: My vision for the next iteration of this functions is a more sophisticated process that uses a little regex to look for all the common phone number formates e.g +1(xxx) xxx-xxx, xxx-xxx-xxxx, etc... and parse accordingly. I don't think we need to worry about non-US yet. Patients count as cases in the state/country of their official residence not where they are infected/test positive.

src/lib/phdi-transforms/phdi_transforms/basic.py
DanPaseltiner: Fair enough. We can change this.

src/lib/phdi-transforms/phdi_transforms/basic.py
graham-skylight: so what do we want to do for right now? Should we see what number formats look like in the actual data?

bamader: I would say we're safe to merge in the 10-digit enforcement as is for the moment because we're not shipping anything yet, and then we write a 1 point fast follow ticket to investigate phone numbers in the real data and modify the logic here accordingly. Remember, our motto is "make it work draft, then iterate and improve"!

PR #51, opened 2022-03-17
Adding Ingest Pipeline

NatHillardUSDS:
This is for https://app.clickup.com/t/2hbh2h7 - to fully automate the ingest pipeline, it is necessary to alter the decrypt function to take input and output args. 

I have successfully used this function as part of the latest Azure Data Factory run to pull data from VA


DanPaseltiner:
Choosing to accomplish this in a different manner start with #110

PR #53, opened 2022-03-18
Add dev version of linkage

bamader:
This PR uploads a _dev version_ of some basic record linkage code (i.e. this is NOT intended to be merge ready--this PR is to get the work shared with the team to offer feedback and so that others can experiment if they desire). Using a representative set of synthetic data (cleared for sharing over slack and github, not that we'll actually need to check it into the repo), this script builds a small fake data set for de-duplication testing and then uses some native pandas to parse the ndjson and perform de-duplicating linkage. I worked off the result of the conversation yesterday with Bryan that we should squash records together where possible, and when there are information conflicts, we preserve all information in the form of a `|` delimited list. So, for feedback, looking at the following:

1. How do folks feel about this style of linkage? Squashing duplicates down into a single "master row" which holds all the concatenated information across records, using strict/exact enforcement only, using pandas natives where possible, etc.
2. How do people feel about the semantics of the linkage aggregator specifically? I maintained any information that wasn't an empty/null that conflicted in the form of a list of unique elements, but this is open to change.
3. What's the best way to get this into the pipeline in the appropriate Azure spot?


src/notebooks/record_dedupe.py
```diff
@@ -0,0 +1,111 @@
...
+    '''
+    Turn a bundle of patient data into a small test set for verifying
+    efficacy of deduplication. There will be 6 unique records here,
+    4 of which are completely dissimilar, one which will have three
+    duplicated versions of varying changes, and one which will be a 
+    near match for the duplicates but is off by one character.
+    '''
+    
+    # Full match case: everything the same, should be no merge conflicts
+    df.loc[6, :] = df.loc[0, :]
```
graham-skylight, line 16: I'm really rusty with pandas - why isn't it `df.loc[6] = df.loc[0]`? Would that do something entirely different?

src/notebooks/record_dedupe.py
```diff
@@ -0,0 +1,111 @@
...
+def create_linking_identifier(df: pd.DataFrame):
+    '''
+    Creates a basic identifier used to check duplicates in an incoming
+    batch of data. The basic identifier for the prototype is
+      FIRSTNAME-LASTNAME-YYYY-MM-DD
+    This identifier is joined to the dataframe proper for use in future
+    computation, as well as potential future use in generating/referencing
+    a master ID.
+    '''
+    df['last_name'] = df['name'].apply(lambda x: x[0]['family'] if type(x[0]['family']) is str else '-'.join(x[0]['family']))
```
graham-skylight, line 44: $0.02: would it be simpler to collate to a list? Something like `df['match-id'] = (df['name']['family'], df['name']['given'], ...)`

src/notebooks/record_dedupe.py
```diff
@@ -0,0 +1,111 @@
...
+        dupe_field: str,
+        group_func: Callable
+    ):
+    '''
+    Perform de-duplication linkage on a dataframe of patient records. 
+    All records that have an *EXACT* match in dupe_field are linked
+    according to the provided group_func. Grouped records are consolidated
+    into single rows, and the de-duplicated dataframe is returned.
+    '''
+    df2 = df.groupby(dupe_field).agg(group_func).reset_index()
```
graham-skylight, line 97: :panda_face: :shrug:

graham-skylight (commented):
Looks good to kick the tires! I'm not totally sure where this fits in overall, and two other things came to mind:

1. Do any of the real records have anything in the `identifier` section? It seems like that's kind of purpose built to track this type of thing
2. Should we generate a block in the identifier section as a surrogate id to match records? That seems like the "proper" way to do it, but I'm not 100% about any knock-on effects there.

src/notebooks/record_dedupe.py
DanPaseltiner: We should find a FHIR compliant way to store the new identifier.

DanPaseltiner: Related to this to should the new identifier be added to the other resources belonging to this patient. This is not possible with data currently loaded, but probably worth considering before import to FHIR server.

src/notebooks/record_dedupe.py
```diff
@@ -0,0 +1,111 @@
...
+    df.loc[8, 'address'] = df.loc[10, 'address']
+
+    # Near-match case: not actually a dupe, but off by one char
+    df.loc[9, :] = df.loc[0, :]
+    dob_list = df.loc[9, 'birthDate'].split('-')
+    dob_list[1] = str((int(dob_list[1]) + 1) % 12)
+    df.loc[9, 'birthDate'] = '-'.join(dob_list)
+
+    return df[:10]
+
```
DanPaseltiner, line 33: In general I'm not a huge fan of selecting columns by their index because if there the order of columns in the data ever changes everything breaks. Using columns names can still break, but it slightly more robust. Not a big deal here given that this func is just making test data for us.

src/notebooks/record_dedupe.py
DanPaseltiner: If we are going to eliminate duplicate records like this then we must update the patient IDs in all resource that link back to the patient resources we are removing.

DanPaseltiner (commented):
I think this is a very good start to baseline record linkage. Before merging I think we should implement the part of record linkage that involves interacting with non-patient resources.

src/notebooks/record_dedupe.py
bamader: Sure. In my experience, it's best to be as absolutely precise as possible when working with `.loc`. One of the things I hate the most (and one of the most insidious ways a system can break) is when  your syntax causes pandas to hand you a `view` or a `reference` into a dataframe rather than the indexed data itself, or a `Series` instead of a partial dataframe. Specifying the nth row and explicitly asking for the _data_ in all columns ensures this is robust to actually fetching data and not a view. Granted, it's in a function that makes dummy test data and so doesn't matter with prod, but still, it's an important practice in my eyes (even if just so I don't slip on other stuff).

src/notebooks/record_dedupe.py
bamader: This has now been addressed with the updated PR. We no longer treat the concatenation of separate fields as the `master_id` (which is beyond the scope of this PR at present, though we will likely need to do a follow up to generate uuid at some point). Moreover, I've added a function that uses the found duplicates in a patient batch to alter the ID pointers in any provided resource to point to the de-duped "master" version of a coalesced record. It's inherently FHIR compliant because, in accordance with our discussion, we're just overwriting make believe numbers anyway. No risk of putting stuff in that the server won't recognize, while also letting us keep a quick and efficient linking system.

src/notebooks/record_dedupe.py
bamader: Dan, yep this is now handled (see above). Graham, lol I too was like I guess 🐼 . Took me like an hour of staring it at the other day to convince myself this was actually the perfect function setup for enforcing exact matching. Groupby returns an iterable with two fields `id, group`, where the `id` parameter is the specific `linking identifier` shared by one or more rows, and `group` is a helpfully named `group` object (basically a Series) that contains views to all the rows defined by `id`. Then, the `.agg` function aggregates all these distinct series into an actual dataframe by doing some string manipulations on their values, throwing out the iterable `id` field, and stacking them on top of each other to produce normal data frame rows.

src/notebooks/record_dedupe.py
bamader: Small note that none of the numbers in this function select columns. The first field of a `.loc[]` access corresponds to the index of a dataframe, which is almost always the row. Column names are specified here by name, which at least in my experience is normal convention (curious why you only regard it as "slightly" more robust? It seems to be the most widely used method of accessing columns, particularly because specifying particular column schemas when people enter data is also par for the course; unless you want to implement fuzzy matching on just column names, which seems like an absolute _nightmare_ 🤮 )

DanPaseltiner: ooops I read the indexing too quickly. It is definitely the most robust thing we can do, but it breaks when column names changes even though it dose not depend on column order which indexing columns dose. I have had problems in the past where there is some template csv that is suppose to be used, but then folks screw with things. This probably won't be an issue for us given the initial conversion to FHIR.

bryan-skylight:
The 093b03d commit brought in a bunch of changes to unrelated files, and appears to mostly just delete them. Please correct this issue and push your code again.

src/FunctionApps/SilverTransforms/RecordLinkage/linkage.py
```diff
@@ -0,0 +1,124 @@
...
+        in X and all other records are blank in X, then the value
+        of the matching columns 1, ..., j is used as a singleton
+      - if some number of non-empty in X records 1, ..., j <= n have 
+        different values in X, then all values 1(X), ..., j(X) are
+        concatenated into a unique list of values delimited by |
+    '''
+    if x.isnull().values.all():
+        return ''
+    non_nans = [x for x in list(x.astype(str)) if x != '']
+    return '|'.join(set(non_nans)) if len(non_nans) > 1 else non_nans[0]
```
bryan-skylight, line 92: The `if ... else ...` logic here is redundant as `.join` returns the single element if there is only one by default.

Given the statement above, this point is not relevant to this specific case, but still worth pointing out to help improve coding logic. The intent of the `if ... else ...` logic here was to concatenate the results with "|" if there was more than one value, however, you use `set` *after* checking the length. Assuming that `non_nans` was equal to `["male", "male"]`, `len(non_nans)` would be greater than 1, but the `len(set(non_nans))` would only be 1. As a result, the logic that has been implemented wouldn't work as intended. Again, because of how `.join` works this won't be an issue, but I wanted to point it out as a learning opportunity.

bryan-skylight (changes requested):
I'm going to stop my review here and send it back as I don't want to overwhelm you with requests. Once the requested changes have been implemented, we can revisit.

src/FunctionApps/SilverTransforms/RecordLinkage/linkage.py
```diff
@@ -0,0 +1,124 @@
...
+      - if some number of records 1, ..., j <= n have the same value 
+        in X and all other records are blank in X, then the value
+        of the matching columns 1, ..., j is used as a singleton
+      - if some number of non-empty in X records 1, ..., j <= n have 
+        different values in X, then all values 1(X), ..., j(X) are
+        concatenated into a unique list of values delimited by |
+    '''
+    if x.isnull().values.all():
+        return ''
+    non_nans = [x for x in list(x.astype(str)) if x != '']
```
bryan-skylight, line 91: I'm wondering if `x.astype(str).tolist()` would be safed to ensure we don't get something like `["m", "a", "l", "e", ...]`

src/FunctionApps/SilverTransforms/RecordLinkage/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+    dupes = find_matches(df, 'linking_identifier')
+    print(dupes)
+
+    df = merge_duplicates(df, 'linking_identifier', record_combination_func)
+    # df.to_csv('./src/notebooks/out.csv', index=False)
+
+    odf = pd.read_json(PATH_TO_RESOURCE_DATA, lines=True)
+
+    contained_ids = list(df['id'])
+    odf['pid'] = odf['patient'].apply(lambda x: x['reference'].split('/')[1].strip())
```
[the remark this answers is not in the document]
bamader: Had to get craft with this guy -- as we wrote it in our pairing, we get a type error from invoking `.str` on non-string elements. Additionally, everything parsed after the `.replace()` invocation became `NaN`, which is obviously not great. So based on a fix we talked about elsewhere for optimization, I dumped the apply by just extracting the column to a list and using list comprehension to pull out the reference IDs.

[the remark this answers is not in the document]
bamader: Update: on discovering the `.map` function, I'm using this instead. It's entire purpose is to vectorize the replacement of values in a column with some other derived value (from either a dictionary or a function). It also times as faster than both `.apply` and `.str` accessing, so seems like an all around win!

DanPaseltiner:
Adding a hash to patient bundles was previously incorporated into main. More sophisticated linkage options will be considered in the future.

PR #55, opened 2022-03-18
Refactor silver transforms to use the library

graham-skylight:


DanPaseltiner (approved):
LGTM - I think this refactor will really serve us well down the line and puts us in a place to support our "building blocks" being used on-prem of with any cloud provider.

PR #56, opened 2022-03-24
[CU-208d8gd] Proof of Concept Notebooks with Spatial Analytics

VincentLaUSDS:
Proof of concept spatial analytics based on use cases informed by [interviews](https://docs.google.com/document/d/1l3HAq-skWRRF_lm8Cf659uvvx7lTbBUm526-RsAse10/edit#) with Fairfax County epidemiologists. This notebook specifically tackles a use case around mapping breakthrough infection rates by Census Block Groups. In our interviews with Fairfax County epis, they effectively showed us many slightly different variations of this visualization with different metrics (e.g. vaccination rates) over different graphical dimensions (e.g. Census Tracts, School Catchment Areas).


VincentLaUSDS:
Screenshot of the first use case of mapping breakthrough infection rates by CBG:

image

src/notebooks/Prototype Spatial Analytics.ipynb
```diff
@@ -0,0 +1,1254 @@
...
+    "        1. personid (Primary Key)\n",
+    "        1. first_name\n",
+    "        1. last_name\n",
+    "        1. street_address\n",
+    "        1. zip_code\n",
+    "        1. latitude\n",
+    "        1. longitude\n",
+    "        1. census_tract\n",
+    "        1. census_block_group\n",
+    "        1. census_block\n",
```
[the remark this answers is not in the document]
VincentLaUSDS: Basically, what I'm trying to say here is that on one hand for just the pilot, I feel like it's fine to just throw a wrapper around SmartyStreets and do geocoding. Getting accurate and precise lat/longs is already a value add.

But then if you were to be really pessimistic you might say "well this is just a wrapper around SmartyStreets, this is kind of dumb". But I think a really cool value add would be if we could also "package" in many other commonly used socioeconomic indices that people often have to manually join in at the census geographical level. At a very high level, I'm thinking that we're not only building a geocoder but also a library of indices that users then don't have to go and search separately and manually join in.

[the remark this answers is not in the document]
graham-skylight: https://www.smarty.com/docs/cloud/us-street-api#http-response-status should be everything we get back from SmartyStreets - worth noting is that I don't see lots of the census stuff you're referencing. We could probably import it into Postgis or similar, but that's a bit of a bigger project

src/notebooks/phdi_analytics.py
```diff
@@ -0,0 +1,297 @@
...
+from concurrent.futures import ThreadPoolExecutor
+from faker import Faker
+from geopy import Point
+from geopy.distance import geodesic
+import geopandas as gpd
+import git
+import matplotlib.ticker as mtick
+import matplotlib.pyplot as plt
+from mpl_toolkits.axes_grid1 import make_axes_locatable
+import names
```
bryan-skylight, line 17: This module is imported and used to generate first and last names, but those names are not used in the analytics. I think we should remove this dependency.

src/notebooks/phdi_analytics.py
```diff
@@ -0,0 +1,297 @@
...
+import numpy as np
+import pandas as pd
+import seaborn as sns
+from tqdm.notebook import tqdm
+
+
+def get_git_root(path):
```
bryan-skylight, line 24: This seems like overkill for a prototype notebook that will only ever reference one Git repo. I'd recommend removing this function, removing the `git` module dependency, and simply hardcoding the path.

src/notebooks/phdi_analytics.py
```diff
@@ -0,0 +1,297 @@
...
+def generate_covid_positive(row):
+    if row['is_fully_vax'] == 0:
+        covid_positive = random.choices([0, 1], weights=(0.97, 0.03), k=1)
+        return covid_positive[0]
+    else:
+        covid_positive = random.choices([0, 1], weights=(0.99, 0.01), k=1)
+        return covid_positive[0]
+
+
+def create_dataframe(radius, number_of_points, center):
```
bryan-skylight, line 132: In production, this function would need to be broken down quite a bit. For example, the generation of the geocoding data (lines 170-177) should be its own function, and this function shouldn't write to a file without some kind of parameter to control that. It's fine for prototype purposes, though; I just wanted to point it out to help reinforce best practices.

src/notebooks/phdi_analytics.py
```diff
@@ -0,0 +1,297 @@
...
+    df['rolling_count'] = df.rolling(window, on='event_dt')[metric].count()
+    df['rolling_rate'] = df['rolling_sum'] / df['rolling_count']
+
+    return df
+
+
+def plot_time_series_metric(df, window, metric_label):
+    df_agg = aggregate_df(df, window=window, metric_label=metric_label)
+
+    window_length_label = re.findall(r'\d+', window)[0]
```
bryan-skylight, line 218: The use of regex here is unnecessary. You should either use `if else` statements to check the values of `window`, or ensure that `window` is an integer and assume that it means number of days. Either way, using regex here gives a big code smell.

src/notebooks/phdi_analytics.py
```diff
@@ -0,0 +1,297 @@
...
+    plt.tight_layout(h_pad=6)
+
+    return fig, ax_lst
+
+
+def create_map(geography, shapefile_path, metric, df):
+    """
+    Create spatial analytics map
+    """
+    shapefile_map = gpd.read_file(shapefile_path)
```
bryan-skylight, line 248: The dependency on `geopandas` is a nightmare on Windows. In order to use `geopandas`, you have to have the `GDAL` API installed, and in order to install that, you have to have the Microsoft Visual C++ 14.0 or later build tools installed. Is there another way to read shapefiles? I remember dealing with these issues in R back when I was doing spatial analytics and it was a nightmare. I'm hoping an alternative has come out that helps alleviate this. If not, no worries.

src/notebooks/phdi_analytics.py
```diff
@@ -0,0 +1,297 @@
...
+    first_names = []
+    last_names = []
+    addresses = []
+    fake = Faker()
+    zip_codes = []
+    event_dates = []
+    for p in points:
+        latitudes.append(p[0])
+        longitudes.append(p[1])
+        address = fake.address()
```
bryan-skylight, line 151: `Faker` is used to create the addresses, but the addresses are never used since the `lat`, `lon` pair are generated programmatically. I recommend getting rid of the `Faker` dependency, and dropping the address generation code.

bryan-skylight (changes requested):
Overall, I like this. But the `phdi_analytics.py` file is trying to do too much. Keep it constrained to only the functionality needed to produce the Prototype Spatial Analytics notebook.

src/notebooks/phdi_analytics.py
VincentLaUSDS: Yeah agreed. I am on Windows, but I'm actually using Anaconda to manage my environment (which I know is inconsistent with the rest of the engineering team and prototype). For this spatial analytics notebook, I think using `geopandas` is the way to go because it really is the simplest way IMO to work with shapefiles in Python / Pandas

bryan-skylight: Agreed. You've got it up and running and it produces the results we need. Plus GDAL seems to be the standard for spatial analytics. But wow, what a nightmare on Windows.

bryan-skylight (approved):
Thanks for implementing those changes. This looks good to me. Nice work, and thanks for putting this together!

PR #59, opened 2022-03-27
ADR for superseding previous ADRs

bryan-skylight:
# Description

This ADR proposes a method by which existing ADRs are deprecated and superseded by new ADRs.

# Feedback Requested

I like the approach in the alternative options section more than the proposed solution, but the proposed solution is closer to the spirt of ADRs. I would be interested in hearing others' thoughts about the two approaches.

### DIFF
--- a/docs/decisions/0005-adr-version-control.md
+++ b/docs/decisions/0005-adr-version-control.md
@@ -1,0 +1,24 @@
+# 5. ADR Version Control
+
+Date: 2022-03-27
+
+## Status
+
+Proposed
+
+## Context and Problem Statement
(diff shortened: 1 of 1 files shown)


JSkeets (commented):
I think mutable ADR would be more difficult to search in the future. I think of it as Jira tickets, If the tickets are linked properly, as proposed here, I can easily search through the history of a task -- in my mind this is the immutable approach.

The mutable approach, feels to me that it would be harder to search, but cut down on clutter. I view searchability as more important than clutter creation. Just my opinion though...

PR #60, opened 2022-03-27
Fix title and date format for ADR #4

bryan-skylight:
# Description

I know ADRs should be immutable once approved or rejected, but this ADR was approved despite inconsistent formatting. Typically I'd let that go, but the title didn't even match the ADR given that the title was `0003 Working with FHIR Data` and this is ADR #4. Given that this needs to be changed, I'm taking the opportunity to also modify the date format to be consistent with the other ADRs.


bamader (approved):

DanPaseltiner (approved):
LGTM

PR #61, opened 2022-03-27
Explicitly state data format in ADR and make Appendix optional

bryan-skylight:
# Description
This PR updates the ADR template to explicitly provide a data format (YYYY-MM-DD) and list the Appendix section as optional. Of the 4 ADRs that have merged to date, 3 of the 4 followed these standards, but ADR 4 showed some variability, so I'm making these updates to make it more explicit what we're looking for in ADRs.


JSkeets (approved):
LGTM

PR #62, opened 2022-03-27
Add Squash and Merge ADR

bryan-skylight:
# Description
This PR adds an ADR explaining our decision to use the Squash and Merge approach to merging PRs into main.

# Feedback Requested
- Does everyone still agree with this approach? It was raised in a conversation by Graham and I believe only a couple people weighed in.  
- Do we need to explicitly describe any type of standards or expectations around the single commit message that is used for the merge?
- Are there other alternatives that should be listed as considered?

### DIFF
--- a/docs/decisions/0006-squash-merge-prs.md
+++ b/docs/decisions/0006-squash-merge-prs.md
@@ -4,7 +4,7 @@ Date: 2022-03-27
 
 ## Status
 
-Proposed
+Accepted
 
 ## Context and Problem Statement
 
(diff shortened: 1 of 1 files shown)


bamader (approved):
I have no strong preferences with regards to either forcing a squash or not, so I defer to others (my previous team didn't squash). I would however _strongly_ support an automation default in the workspace that automatically squashes branch commits when they're merged to base, rather than requiring the developer to locally do squashing down.

PR #63, opened 2022-03-28
Add ADR requiring checks to pass before approvals are made

bryan-skylight:
# Description
Pull Requests require both that all checks that have been implemented pass and that at least 1 developer approve the pull request. Unfortunately, there's no way to ensure that a developer is only approving pull requests that have passed all checks, which can result in an opportunity for developers to merge PRs with unsuccessful checks. This PR introduces two ADRs, which encompass two different solutions to this. I assume that one of the ADRs will be rejected in favor of the other, but I wanted to provide both as an option.

# Feedback Requested
- Which of these seems like the best approach to ensuring bad merges don't take place?  
- Is there another option I'm not considering?

### DIFF
--- a/docs/decisions/0007-pr-approval-requires-checks-to-pass.md
+++ b/docs/decisions/0007-pr-approval-requires-checks-to-pass.md
@@ -18,7 +18,7 @@ One of the conditions that must be met before a Pull Request can be merged into 
 
 ## Considered Options
 
-No other options were considered.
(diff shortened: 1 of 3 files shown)


JSkeets:
I vote for PRs are blocked until all status checks are completed. If we were supporting a production application there would be reason to allow for the option to override these to get hotfixes out, but this doesn't apply to us.

JSkeets (approved):
I vote for PRs are blocked until status checks complete

bamader (approved):
I would vote for merge is blocked until checks are passing but _not_ block approving by checks not passing. I could be viewing the approval function differently than other folks, but to me an approval represents a sign-off that the code and the ideas it encapsulates are sound. If we can't merge until checks pass, I don't see what extra value is added by blocking approvals as well.

PR #66, opened 2022-03-30
[BULK-IMPORT] Import ndjson into FHIR

JSkeets:
# Description
This pull request creates the code which performs the import of a zip file into a FHIR server. It does so by

- Unzipped the file
- Creating parallel processes for each file in the unzipped directory
- Reading each `ndjson` line within the directory
- Posting that line to the FHIR server

The greater orchestration of this code into the function app resources as whole will greatly determine some aspects of this code, such as how it is called, etc. I operated under discussion that a zip file would be passed to this function.

I recognize that there will be details that have to be sorted with this code once the deployment of it as one function app and part of a pipeline begin, but I feel that this code does its job of uploading resources from a zip file into the FHIR server, and the orchestration piece is TBD

# Testing

This function has been tested by using synthea data in a zip file and receiving a positive response from the FHIR server, verifying that the patient exist

Screen Shot 2022-03-30 at 12 04 05 PM
Screen Shot 2022-03-30 at 11 39 55 AM


src/FunctionApps/BulkImport/Import/main.py
```diff
@@ -0,0 +1,98 @@
...
+import os
+import zipfile
+
+import requests
+import multiprocessing
+import json
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
+
+unzipped_directory = "./FhirResources"
```
bamader, line 11: (non-blocking): We'll have to change the input assumption for this function once it hits the Azure environment. For purposes of orchestration, if the plan is to write a master function that calls the other functions with direct parameter handoff, this is easy to accomplish, but if we're going to go with more of an API route, then this gets trickier. Potentially something to talk through soon (though I guess we'll have to see tickets come out of somewhere first).

src/FunctionApps/BulkImport/Import/main.py
```diff
@@ -0,0 +1,98 @@
...
+
+def main(zip_file):
+    unzip_input_file(zip_file)
+    process_ndjson_files(unzipped_directory)
+
+def unzip_input_file(zip_file):
```
bamader, line 17: I think we'll ultimately want doc strings for each function that give types for both inputs and outputs, as well as provide a summary of the method's invocation. e.g.

```
def unzip_input_file(zip_file: str) -> str:
    '''
    Opens the given file and unzips to a new directory for processing.
    '''
    [code here]
```

src/FunctionApps/BulkImport/Import/main.py
```diff
@@ -0,0 +1,98 @@
...
+    with zipfile.ZipFile(zip_file,"r") as zip_ref:
+        zip_ref.extractall(unzipped_directory)
+
+def process_ndjson_files(dir):
+    with multiprocessing.Pool(processes = multiprocessing.cpu_count()-1) as p:
+        directory = os.fsencode(dir)
+        for file in os.listdir(directory):
+            filename = os.fsdecode(file)
```
bamader, line 25: Can we improve clarity on these variable names? I can't tell at a glance what directory `directory` refers to (i.e. I thought someone handed you a zip file, not a directory). Ditto for filename.

bamader: Though on second thought this might also change depending on the orchestration implementation so potentially nvm

src/FunctionApps/BulkImport/Import/main.py
```diff
@@ -0,0 +1,98 @@
...
+        p.close()
+        p.join()
+
+def read_file(file):
+    with open(file) as fp:
+        for line in fp:
+            json_line = json.loads(line)
+            resource_type = json_line["resourceType"]
+            resource_id   = json_line["id"]
+            post_to_fhir(line,resource_type,resource_id)
```
bamader, line 40: Out of curiosity, what's the performance on this? I know the FHIR server has no bulk import (which absolutely sucks 🙄 ), but do we need to speed up line by line json reading? I can't imagine it's that slow so maybe a non issue.

src/FunctionApps/BulkImport/Import/main.py
```diff
@@ -0,0 +1,98 @@
...
+    try:
+        token = get_access_token()
+    except Exception:
+        return func.HttpResponse("error getting access token", status_code=401) 
+    retry_strategy = Retry(
+    total=3,
+    status_forcelist=[429, 500, 502, 503, 504],
+    allowed_methods=["HEAD", "POST", "OPTIONS"]
+    )
+    adapter = HTTPAdapter(max_retries=retry_strategy)
```
bamader, line 52: Is this an access token for the FHIR server? Or to be able to run the function in the environment?

JSkeets: This is the access token, once we orchestrate things will change as we have the new auth method.

src/FunctionApps/BulkImport/Import/main.py
JSkeets: The parallelizing really helped, I am not sure how many lines we have to process for the ultimate upload, I also have a very old slow macbook which doesn't help. I think we won't know for sure til we get to testing on a function app

PR #68, opened 2022-03-31
RFC: Intake pipeline skeleton app

graham-skylight:


src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -0,0 +1,27 @@
+import logging
+
+import azure.functions as func
+
+from IntakePipeline.transform import transform_bundle
+from IntakePipeline.linkage import add_patient_identifier
+from IntakePipeline.fhir import read_fhir_bundles, upload_bundle_to_fhir_server
```
bamader, line 7: Not a blocker, just for clarification: this implies that we'll move all the other function code into a single giant function within the function app. It'll be the Python function app, which has the IntakePipeline function directory, and inside of that, three sub directories, one for each of the `transform`, `linkage`, and `fhir` code. Some q's:
1. how do folks feel about having one function within the function app? I personally have no issue with it, given that our whole goal here was to create a simplified single master invocation function that called everything else. 
2. if we did move to a structure that had one function and three subdirectories (I think it'll go to 4 once we merge in Jesse's import code), does that mean we should dump the `__init__.py` and various function files for each of the subdirectories? I.e. will we have only a single `__init__.py` here in Intake Pipeline?

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def run_pipeline():
+    for bundle in read_fhir_bundles():
+        transform_bundle(bundle)
+        add_patient_identifier(bundle)
+        upload_bundle_to_fhir_server(bundle)
```
bamader, line 14: Again, I like this. Point to note here is that Graham found that the real VA data is broken up into a massive set of smaller files, each around ~20 mb. That's _key_ for us (or at least finding that out was key to me), because it actually solves our memory bottleneck problem. If the FHIR server only converts one bundle (= one file = 20 mb) at a time, then our Function App will never run into that 1.5 gb limit. There will be another step at the top of this `run_pipelin()` function before we `transform`, where we run `convert_bundle_by_message_to_bundle_by_resource`, and we might wind up doing some data returns in between, (e.g.
```
def run_pipeline():
    for bundle in read_fhir_bundles():
        resource_packet = convert_bundle_by_message_to_bundle_by_resource(bundle)
        resource_packet = transform_bundle(resource_packet)
        resource_packet = add_linking_identifier(resource_packet)
        resource_packet = link_duplicate_records(resource_packet)
        upload_bundle_to_fhir_server(resource_packet)
```
), but overall this is exactly what I had in mind.

src/FunctionApps/python/IntakePipeline/fhir.py
```diff
@@ -0,0 +1,9 @@
+from typing import Iterator
+
+
+def read_fhir_bundles() -> Iterator[dict]:
+    yield {}
```
bamader, line 5: Graham, is the intention for this function to provide an iterator over the actual converted FHIR bundles? My yield fu is weak right now w.r.t. to memory implications (I _think_ yield doesn't load anything until the calling for loop asks for the object), but will this work with our memory constraints? Or do we need to make this an iterator over file names and then open individual files in another function?

src/FunctionApps/python/IntakePipeline/fhir.py
```diff
@@ -0,0 +1,9 @@
...
+
+
+def upload_bundle_to_fhir_server(bundle: dict) -> None:
```
bamader, line 8: If a `bundle` is a bunch of HL7 messages that was converted to FHIR, won't that put ndjson rather than just a json/dict object? I could be wrong, but that parameter might need to be a list of dicts.

src/FunctionApps/python/IntakePipeline/__init__.py
bamader: On second thought, we could theoretically keep our individual functions in their own directories right now, with their own `__init__.py` files and associated tests, and then just make code copies of the things that need to be ported to work with the pipeline. Then we could just deploy the `IntakePipeline` to Azure as a one-stop shop.

src/FunctionApps/python/IntakePipeline/linkage.py
```diff
@@ -0,0 +1,2 @@
+def add_patient_identifier(bundle: dict) -> None:
+    pass
```
bamader, line 2: There'll obviously be more functions to fill in here for each of the sub-directories, but I like that theoretically, this orchestration skeleton lets us parallelize the work around fleshing out each of the functions that goes into making the pipeline run. As long as each of us is working on a different one of these sub-files, and as long as we document what the inputs/outputs to each one w.r.t. to the main pipeline are, then we should be able to refine all of the functions simultaneously.

bamader (approved):
Overall this looks good to me. Have a few Qs/discussion points in the comments that I'd love other folks' thoughts on, but overall I think this enables us to move forward as intended. One thing we will have to decide is how to implement the inter-batch record linkage (I spent some time spiking the FHIR server and I think I've found that it doesn't perform any kind of reasonable linkage to help us, which means we'll need a separate cache or DB or table or something to query patients we've already seen). But that problem can be confined more towards the ticket around making linkage work with orchestration.

src/FunctionApps/python/IntakePipeline/__init__.py
graham-skylight: One of the big ideas here is to keep the bundles intact for now - Azure's FHIR bulk import doesn't work, so there's no requirement to group them.

src/FunctionApps/python/IntakePipeline/fhir.py
graham-skylight: basically - the idea here is that downstream, people just for a `for x in y:` and don't have to worry about it. In the background, yep, it's about memory efficiency.

src/FunctionApps/python/IntakePipeline/fhir.py
graham-skylight: nope - it's not ndjson, it's just json

DanPaseltiner (approved):
At a high level this looks go to me. I think it gives us what we need to parallelized work to the best of our ability.

src/FunctionApps/python/IntakePipeline/__init__.py
bamader: For sure, no requirement to regroup them but we do want to do that for the record linkage function. That's how intra-batch linkage finds duplicate patients and that's how we can iterate over non-patients to change the ID they point to. It's just really helpful to have things split up like that, and since Dan already has the function (or 80% of one) hopefully it's a relatively low effort to plug in.

src/FunctionApps/python/IntakePipeline/fhir.py
bamader: Huh. So what does that json look like? If a bundle of HL7 messages is just a list of dicts conforming to some specification, the `convert_to_fhir` function should turn that into a list of dicts conforming to a different standard as I understood it. If it becomes json, how does that work with keys and mappings?

spencer-skylight: Yeah, ndjson is a different concept from Bundles.  A bundle is actually its own FHIR resource type that includes a collection of other resources.  

Here's an example of a FHIR bundle:
https://www.hl7.org/fhir/diagnosticreport-example-f001-bloodexam.json.html
and the spec...
https://www.hl7.org/fhir/bundle.html

spencer-skylight (approved):
Looks good to me.

src/FunctionApps/python/IntakePipeline/__init__.py
spencer-skylight: I'm curious...  In what way does Azure's bulk import not work?  What has been attempted or what documentation is out there that does not work properly?

graham-skylight: Jesse got `The bulk import feature is currently available in the open source FHIR server. It's not available in Azure Health Data Services yet.` a couple weeks ago on the $import endpoint

PR #69, opened 2022-04-04
Graham/jsonify dstp

DanPaseltiner:
Previously the files returned by convertToFHIR() were a hybrid of nicely formatted JSON (containing line breaks and indents) and NDJSON. Despite being human-readable these files could not be read programmatically line by line. This change removes all line breaks within the string returned from convertMessageToFHIR(), but continues to append these strings together with line feeds. The result are files that contain the entire FHIR bundle from single HL7 message on one line. Our ability to read these files has been tested with Python's `ndjson.loads(<file-as-string>) and json.loads(<one-line-of-file>)`.


bamader (approved):
Good change and even better catch. This should be especially memory friendly to just stick at the top of the `IntakePipeline` main execution code.

graham-skylight (approved):
:+1: for now but we really need to rewrite (either in python, or with tests)

PR #70, opened 2022-04-04
Start ingesting data from container files

graham-skylight:
Pulls files from containers and begins to yield them into the pipeline. Currently works on ndjson files (which should populate the `valid-messages` container after #69 )


src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -7,8 +8,18 @@
 from IntakePipeline.fhir import read_fhir_bundles, upload_bundle_to_fhir_server
 
 
+def get_required_config(varname: str) -> str:
+    """Grab a required config val and throw an exception if it's not present"""
+    if varname not in os.environ:
+        raise Exception(f"Config value {varname} not found in the environment")
+    return os.environ[varname]
```
bamader, line 15: Wound up copying this for my `add_patient_identifier` code. Works like a charm. I moved it into a `utils.py` file since I wanted to access it in some different ways, but otherwise absolutely no issues with it.

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -7,8 +8,18 @@
...
+
+
 def run_pipeline():
-    for bundle in read_fhir_bundles():
+    for bundle in read_fhir_bundles(
+        get_required_config("INTAKE_CONTAINER_URL"),
+        get_required_config("INTAKE_CONTAINER_PREFIX"),
```
bamader, line 21: Is it worth just storing these two values in strings above the for loop, so that we don't have to make a pair of function calls with every bundle that passes through? e.g.

```
intake_container_url = get_required_config()
intake_container_prefix = get_required_config()
for bundle in read_fhir_bundles(intake_container_url, intake_container_prefix):
    pass
```

I stored the salt hash that way.

src/FunctionApps/python/IntakePipeline/fhir.py
```diff
@@ -1,8 +1,36 @@
...
 
+from typing import Iterator, IO
 
-def read_fhir_bundles() -> Iterator[dict]:
-    yield {}
+from azure.identity import DefaultAzureCredential
+from azure.storage.blob import ContainerClient
+
+
+def get_blob_client(container_url: str) -> ContainerClient:
```
bamader, line 11: We'll probably want some function descriptor docstrings in all of these functions (i.e

```
def get_blob_client(container_url: str) -> ContainerClient:
'''
Function to blah blah blah
'''
    pass
```

src/FunctionApps/python/IntakePipeline/README.md
```diff
@@ -0,0 +1,9 @@
+# Intake Pipeline Function App
+
+## Config
+
+This function app requires some config to be present. In production this is defined on the function app, and locally, you'll want to add
+items in `local.settings.json` under `Values`.
+
+* `INTAKE_CONTAINER_URL`: a container URL containing ndjson formatted FHIR records (eg: https://pitestdatasa.blob.core.windows.net/bronze)
+* `INTAKE_CONTAINER_PREFIX`: a string prefix so we're not scanning the entire container (eg: decrypted/valid-messages)
```
bamader, line 9: Are there circumstances where we'd want to process only a subset of the files in a container through the pipeline? I'm imagining we'll be pretty meticulous about what files we put in the decrypted containers, so why not just scan the whole container and be careful with file maintenance outside the pipeline?

bamader (approved):
LGTM. I would organize slightly differently and use a `utils.py` file but I don't see any logic errors or type issues or anything like that (and I can always make that change on my branch after this gets merged, if desired). Just some quick commenting and this is good to go imo.

DanPaseltiner (approved):
I agree with Brandon's comment on having a doc string for all functions. Beyond that, LGTM.

src/FunctionApps/python/IntakePipeline/__init__.py
graham-skylight: :shrug: env var reading is quick

src/FunctionApps/python/IntakePipeline/README.md
graham-skylight: that bucket has all the encrypted files as well, so this gives us a little flexibility - also, for example, if we wanted to re-run just ELRs

PR #71, opened 2022-04-05
Fhir docs

spencer-skylight:
Create documentation around performing FHIR-related queries and authentication.

### DIFF
--- a/docs/fhir/fhir-auth.md
+++ b/docs/fhir/fhir-auth.md
@@ -1,0 +1,31 @@
+# Introduction
+This file describes required authorization processing for authorizing to the PHDI FHIR server.
+
+## Client Credentials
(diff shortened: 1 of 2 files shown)


bamader (approved):
Seems legit!

DanPaseltiner (commented):
In general I thought we were trying to keep this type of documentation in the private repo. I could be wrong. Might be worth a conversation before merging.

spencer-skylight:
@DanPaseltiner that makes sense... I won't be on the standup tomorrow to discuss unfortunately, but perhaps the team could discuss and I can update accordingly.

spencer-skylight:
Moved to: https://github.com/usds/PRIME-PHDI-Internal/pull/17

PR #72, opened 2022-04-05
Add patient identifier to pipeline code

bamader:
This pull request fleshes out the code we decided on for the May 3rd deadline to add a unique patient identifier to each patient in a bundle that comes through our pipeline. Per our discussion, this identifier is a salt + hash of a linking string built out of the patient's name, date of birth, and address. We then store this directly back into the patient resource in the `identifiers` array. This changeset also leverages an environment config utility that Graham wrote, although I situated it in a different file with some new organization (code is untouched though). Finally, I've included tests for both of the core functions the linking identifier is built on.


src/FunctionApps/python/IntakePipeline/linkage.py
```diff
@@ -1,2 +1,52 @@
...
+            link_str = name_str + '-' + patient['birthDate'] + '-' + address_line
+            hashcode = generate_hash_str(link_str, salt_suffix)
+            patient['identifier'].append({
+                'value': hashcode,
+                # Note: this system value corresponds to the FHIR specification
+                # for a globally used / generated ID or UUID--the standard here
+                # is to make the use "temporary" even if it's not
+                'system': 'urn:ietf:rfc:3986',
+                'use': 'temp'
+            })
```
DanPaseltiner: Could we add some condition to stop this loop once a patient resource is found? No need to keep going through all the resources after the patient has been found unless we think that a single ELR, VXU, or ECR might have multiple patients.

src/FunctionApps/python/IntakePipeline/linkage.py
```diff
@@ -1,2 +1,55 @@
...
+            # Generate and store unique hash code
+            link_str = name_str + "-" + patient["birthDate"] + "-" + address_line
+            hashcode = generate_hash_str(link_str, salt_suffix)
+            patient["identifier"].append(
+                {
+                    "value": hashcode,
+                    # Note: this system value corresponds to the FHIR specification
+                    # for a globally used / generated ID or UUID--the standard here
+                    # is to make the use "temporary" even if it's not
+                    "system": "urn:ietf:rfc:3986",
```
graham-skylight, line 40: It's not a URI - is there anything that says we can't just make up a system so it doesn't clash? Maybe even something that points back to us specifically?

src/FunctionApps/python/IntakePipeline/linkage.py
```diff
@@ -1,2 +1,55 @@
...
+def generate_hash_str(linking_str: str, salt_str: str) -> str:
+    """
+    Given a string made of concatenated patient information, generate
+    a hash for this string to serve as a "unique" identifier for the
+    patient.
+    """
+    hash_obj = hashlib.sha256()
+    to_encode = (linking_str + salt_str).encode("utf-8")
+    hash_obj.update(to_encode)
+    return str(hash_obj.hexdigest())
```
graham-skylight, line 55: `.hexdigest()`'s return value is already `str`

src/FunctionApps/python/IntakePipeline/linkage.py
bamader: That's actually why I didn't specify a terminating condition. To me it seemed like defensive coding to not stop once we found a patient, given that we don't have a guarantee a bundle _won't_ contain multiple patients. And, especially for our early deadlines, this thing doesn't have to be speed performant at all. It just has to run.

DanPaseltiner: Fair enough!

graham-skylight (changes requested):
Looks good overall, but we may want to be a bit more specific in the fields we're using

src/FunctionApps/python/IntakePipeline/linkage.py
```diff
@@ -1,2 +1,55 @@
-def add_patient_identifier(bundle: dict) -> None:
-    pass
+import hashlib
+
+
+def add_patient_identifier(salt_suffix: str, bundle: dict) -> None:
```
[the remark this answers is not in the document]
bamader: easy enough fix

src/FunctionApps/python/IntakePipeline/linkage.py
```diff
@@ -1,2 +1,55 @@
...
+        patient resource
+    This function assumes data has been standardized already by the
+    silver transforms.
+    """
+    for resource in bundle["entry"]:
+        if resource["resource"]["resourceType"] == "Patient":
+            patient = resource["resource"]
+
+            # Combine given and family name
+            recent_name = patient["name"][0]
```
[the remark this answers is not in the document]
bamader: Mmmm yeah good call

src/FunctionApps/python/IntakePipeline/linkage.py
bamader: I know it's not a URI but that's the FHIR specification for what to put in the system field if you're rolling your own "global" or otherwise custom identifier. If we were to publish an organization page and then fill our some documented information on there we could def point it back at us, but short of that, everything I've read in the docs says to just use this so that others working with the data know we basically made this up.

src/FunctionApps/python/IntakePipeline/linkage.py
```diff
@@ -1,2 +1,55 @@
...
+                    # Note: this system value corresponds to the FHIR specification
+                    # for a globally used / generated ID or UUID--the standard here
+                    # is to make the use "temporary" even if it's not
+                    "system": "urn:ietf:rfc:3986",
+                    "use": "temp",
+                }
+            )
+
+
+def generate_hash_str(linking_str: str, salt_str: str) -> str:
```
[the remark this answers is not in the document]
bamader: Meh I do prefer it explicit on salt at least (because salt on its own imo is not a super clear variable name), but I can change the other to `linking_identifier` since that what it is.

src/FunctionApps/python/IntakePipeline/linkage.py
bamader: For some insight into how FHIR thinks about ID systems, check out their explanation of the system parameter under the identifier table: https://www.hl7.org/fhir/datatypes.html#Identifier

And then here are the applicable types of identifier systems they accept: https://www.hl7.org/fhir/identifier-registry.html. We can't use a URL because that has to resolve, and we don't really fit any other category. And in the context of our application, patients do theoretically have to get unique hashes generated for them. I agree it's not a perfect fit but seems like the best we can do given their specc?

graham-skylight (approved):
:+1:

PR #74, opened 2022-04-05
Implement patient transforms

graham-skylight:
* Name, phone, and geocoding

### DIFF
--- a/src/FunctionApps/python/IntakePipeline/transform.py
+++ b/src/FunctionApps/python/IntakePipeline/transform.py
@@ -1,2 +1,59 @@
-def transform_bundle(bundle: dict) -> None:
-    pass
+from smartystreets_python_sdk import us_street
+
+from phdi_transforms.basic import transform_name, transform_phone
+from phdi_transforms.geo import geocode
+
+
+def find_patient_resources(bundle: dict) -> dict:
+    """Grab patient resources out of the bundle, and return a reference"""
+    return [
+        r
+        for r in bundle.get("entry")
+        if r.get("resource").get("resourceType") == "Patient"
+    ]
+
+
+def transform_bundle(client: us_street.Client, bundle: dict) -> None:
+    """Standardize name and phone, geocode the address"""
+
+    for resource in find_patient_resources(bundle):
+        patient = resource.get("resource")
+
+        # Transform names
+        for name in patient.get("name", []):
+            if "family" in name:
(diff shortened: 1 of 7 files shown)


src/FunctionApps/python/IntakePipeline/transform.py
```diff
@@ -1,2 +1,68 @@
...
+    """Grab a patient resource out of the bundle, and return a reference"""
+
+    patients = [
+        r
+        for r in bundle.get("entry")
+        if r.get("resource").get("resourceType") == "Patient"
+    ]
+
+    if patients:
+        return patients[0]
```
DanPaseltiner, line 17: This won't work if there are multiple patient resources in a bundle. In theory I don't think that should happen, but it is something to be aware of.

src/FunctionApps/python/IntakePipeline/transform.py
```diff
@@ -1,2 +1,68 @@
...
+    if not resource:
+        return None
+
+    patient = resource.get("resource")
+
+    # Transform names
+    for name in patient.get("name", []):
+        if "family" in name:
+            name["family"] = transform_name(name["family"])
+
```
DanPaseltiner, line 33: Is family name only ever a single value and never a list?

DanPaseltiner (approved):

src/FunctionApps/python/IntakePipeline/transform.py
bamader: I would prefer to see us code things to handle all patients in a bundle. While I do agree that theoretically that shouldn't happen, it's a small defensive thing we can do to make our code more robust and handle something I don't think we can/should safely assume.

src/FunctionApps/python/IntakePipeline/transform.py
bamader: I've only ever seen it as a single value. Given name comes as a list, but I haven't seen that for family name. It's something we could do defensively if we wanted but I'm fine doing single value for now.

bamader (changes requested):
Other than accounting for multiple patients in a bundle, this all looks good to me.

src/FunctionApps/python/IntakePipeline/transform.py
graham-skylight: Let's wait until we see it as a list somewhere, no need to get super defensive for the prototype

PR #76, opened 2022-04-06
Initial skeleton for CSV generation function

graham-skylight:


bamader (commented):
The general flow makes sense to me and I think this is a pretty intuitive take on a CSV structure. One potential issue I see is whether there are any circumstances in which a single bundle would necessitate more than one line being written. An example might be a bundle containing more than one patient record (which is imo a good defensive piece we implemented in the earlier steps). If that would be better on multiple rows, we should change the `generate_csvs()` function to use `.writerows()` and alter the return types of the individual functions. That may be a good idea anyway in case we later identify situations of wanting a bundle split into multiple rows. Not a major change by any means, but could be worth discussing.

graham-skylight:
@bamader let's do that after 5/3 or if we actually see it in the data

bamader (approved):

gskathol:
I think this looks like a good framework.  Wondering if it makes sense to at least write data to a dataframe as an intermediate function.  I'm not sure if this would be any harder than going directly to CSV - you would basically get the to_csv for free at the end.  Vincent's Jupyter demo runs off a dataframe, so seems like it would be nice to have if it's not much more effort.

spencer-skylight (approved):
I think this looks like a good framework. Wondering if it makes sense to at least write data to a dataframe as an intermediate function. I'm not sure if this would be any harder than going directly to CSV - you would basically get the to_csv for free at the end. Vincent's Jupyter demo runs off a dataframe, so seems like it would be nice to have if it's not much more effort.

graham-skylight:
@spencer-skylight I don't really see a concrete benefit there right now - maybe down the line a bit?

PR #79, opened 2022-04-11
Misc pipeline fixes

graham-skylight:
* Basically just covering for missing data we didn't anticipate

### DIFF
--- a/src/lib/phdi-transforms/phdi_transforms/geo.py
+++ b/src/lib/phdi-transforms/phdi_transforms/geo.py
@@ -30,7 +30,7 @@ def geocode(client: us_street.Client, address: str) -> GeocodeResult:
     lookup = Lookup(street=address)
     client.send_lookup(lookup)
 
-    if lookup.result:
+    if lookup.result and lookup.result[0].metadata.latitude:
         res = lookup.result[0]
         addr = [res.delivery_line_1]
         if res.delivery_line_2:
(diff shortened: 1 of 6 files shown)


src/lib/phdi-transforms/phdi_transforms/geo.py
```diff
@@ -30,7 +30,7 @@ def geocode(client: us_street.Client, address: str) -> GeocodeResult:
     lookup = Lookup(street=address)
     client.send_lookup(lookup)
 
-    if lookup.result:
+    if lookup.result and lookup.result[0].metadata.latitude:
```
DanPaseltiner, line 33: What's the problem we are trying to solve here? Were we getting back empty results?

DanPaseltiner (approved):
I don't quite follow the change in geo.py, but trust that with the context of the problem it is solving things would make sense. Beyond that this looks good to me. Perhaps post-prototype we should investigate these occurrences of missing data to make sure it isn't some how a result of the FHIR conversion process.

src/lib/phdi-transforms/phdi_transforms/geo.py
graham-skylight: I did a pretty long local run with the other changes, and then got a response where a seemingly valid response had blank lat/lng and crashed out.

JSkeets (approved):
LGTM

src/lib/phdi-transforms/phdi_transforms/geo.py
DanPaseltiner: I figured it was something like that.

PR #80, opened 2022-04-12
refactor config

graham-skylight:
we're probably using the same config mechanisms across function apps, this refactors it out of the `IntakePipeline` function


bamader (approved):
Easy approve 👍

PR #81, opened 2022-04-12
Dan/patient resource parsing

DanPaseltiner:
This PR adds a parse_patient_resource() function that given a patient resource returns a comma deliminated string containing the fields we intend to include in the CSVs that originate from the patient resource.

### DIFF
--- a/src/FunctionApps/python/GenerateCSVs/patient.py
+++ b/src/FunctionApps/python/GenerateCSVs/patient.py
@@ -97,20 +97,17 @@ def get_value(dictionary: dict, key: str) -> str:
 
 
 if __name__ == "__main__":
-    #import json
-    #bundle = json.load(open(r"/Users/danielpaseltiner/Skylight/real-VA-data/00066837-0c33-4526-9ccb-ca750b94794d.fhir"))
-    #breakpoint()
-
     patient = {'fullUrl': 'urn:uuid:c53f9ad8-34c1-ce05-c0f6-7a0ea7bd8483', 
-    'resource': {'resourceType': 'Patient', 
+                'resource': {'resourceType': 'Patient', 
                 'id': 'c53f9ad8-34c1-ce05-c0f6-7a0ea7bd8483', 
                 'identifier': [{'value': 'E42308111'}, {'value': '292561276fcdefab6a2a1545abf7aa9bf30906ba6f0d4f8faff652efc3b4ab3c', 'system': 'urn:ietf:rfc:3986', 'use': 'temp'}], 
-                'name': [{'family': 'FOWLKES', 'given': ['CHERYL'], 'use': 'official'}], 
-                'birthDate': '1960-07-28', 
+                'name': [{'family': 'DOE', 'given': ['JANE'], 'use': 'official'}], 
+                'birthDate': '1950-01-28', 
                 'gender': 'female', 
-                'address': [{'line': ['4903 Falcon Creek Way'], 'city': 'Hampton', 'state': 'VA', 'postalCode': '23666', 'use': 'home', 'extension': [{'url': 'http://hl7.org/fhir/StructureDefinition/geolocation', 'extension': [{'url': 'latitude', 'valueDecimal': 37.09339}, {'url': 'longitude', 'valueDecimal': -76.39775}]}]}], 
+                'address': [{'line': ['123 Main Street'], 'city': 'Gotham', 'state': 'XY', 'postalCode': '12345', 'use': 'home', 'extension': [{'url': 'http://hl7.org/fhir/StructureDefinition/geolocation', 'extension': [{'url': 'latitude', 'valueDecimal': 3.14159}, {'url': 'longitude', 'valueDecimal': 2.71828}]}]}], 
                 'telecom': [{'system': 'phone', 'use': 'home'}], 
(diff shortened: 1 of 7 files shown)


bamader (approved):
This looks great! Makes the rest of the CSV processing substantially easier. All the defensive coding and the length of that test file look extremely not fun though :/

src/FunctionApps/python/GenerateCSVs/patient.py
```diff
@@ -0,0 +1,166 @@
+PATIENT_COLUMNS = [
+    "patient_hash",
+    "given_name",
+    "family_name",
+    "birthDate",
```
graham-skylight, line 5: $0.02: I'm not totally in love with the mixed case - can we pick underscores or camel?

src/FunctionApps/python/GenerateCSVs/patient.py
```diff
@@ -0,0 +1,166 @@
...
+    "state",
+    "postalCode",
+    "latitude",
+    "longitude",
+    "race",
+    "ethnicity",
+]
+
+
+def parse_patient_resource(pt_rsc: dict) -> str:
```
graham-skylight, line 18: Awesome, but it's probably easier to work with downstream if you return `List[Any]` so other functions can jam their `List`s on the end

graham-skylight, line 18: Awesome, but it's probably easier to work with downstream if you return `List[Any]` so other functions can jam their `List`s on the end

src/FunctionApps/python/GenerateCSVs/patient.py
```diff
@@ -0,0 +1,166 @@
...
+    identifiers = pt_rsc["resource"].get("identifier")
+    if identifiers:
+        for id in identifiers:
+            if id.get("system") == "urn:ietf:rfc:3986":
+                hash = id.get("value")
+
+    return hash
+
+
+def get_name(pt_rsc: dict) -> str:
```
graham-skylight, line 49: `-> Tuple[str, str]` or `List[str]` may be easier to work with

graham-skylight, line 49: `-> Tuple[str, str]` or `List[str]` may be easier to work with

src/FunctionApps/python/GenerateCSVs/patient.py
```diff
@@ -0,0 +1,166 @@
...
+
+    race = ""
+    ethnicity = ""
+    if "extension" in pt_rsc["resource"]:
+        for extension in pt_rsc["resource"]["extension"]:
+            if (
+                extension["url"]
+                == "http://hl7.org/fhir/us/core/StructureDefinition/us-core-race"
+            ):
+                race = extension["extension"][0]["valueCoding"]["display"]
```
graham-skylight, line 145: $0.02: do we want to use the `display` values here? Would codes be easier to parse later on?

graham-skylight, line 145: $0.02: do we want to use the `display` values here? Would codes be easier to parse later on?

src/FunctionApps/python/tests/GenerateCSVs/test_patient.py
```diff
@@ -0,0 +1,95 @@
+from GenerateCSVs.patient import parse_patient_resource
+
+
+def test_parse_patient_resource():
+
+    patient = {
+        "fullUrl": "urn:uuid:c53f9ad8-34c1-ce05-c0f6-7a0ea7bd8483",
```
graham-skylight, line 7: Tests look great - like this is totally fine, but if you want to load them from files (like `tests/IntakePipeline/assets/patient_bundle.json`) that'd be cool too

graham-skylight, line 7: Tests look great - like this is totally fine, but if you want to load them from files (like `tests/IntakePipeline/assets/patient_bundle.json`) that'd be cool too

graham-skylight (changes requested):
Looks great - we talked a little about writing the `PATIENT_COLUMNS` and the initial data out to files before calling into the `csv_to_x` functions, but I don't think that's going to get us too much here. The only thing I'd really like to see changed is returning a list instead of a CSV string

src/FunctionApps/python/GenerateCSVs/patient.py
bamader: Dan had mentioned he wanted to keep the variable names as close to the FHIR spec as possible where it could apply. I think that's why you see some mixing (e.g. `birthDate`).

DanPaseltiner: As Brandon mentioned I just kept the names from FHIR. I figured that reducing on the number of different ways refer to birthday (especially when every STLT will have their own conventions) might make things a little easier down the road. Happy to pick names we like more though if others disagree.

graham-skylight: :shrug: I definitely get the point of keeping it in line with FHIR - in that case, let's make the new ones camel as well

src/FunctionApps/python/tests/GenerateCSVs/test_patient.py
DanPaseltiner: Implemented this change.

src/FunctionApps/python/GenerateCSVs/patient.py
DanPaseltiner: That's a good call. We are now grabbing the codes themselves.

src/FunctionApps/python/GenerateCSVs/patient.py
DanPaseltiner: done

graham-skylight (approved):
:tada:

PR #83, opened 2022-04-12
Write csvs to blob storage

graham-skylight:


src/FunctionApps/python/GenerateCSVs/__init__.py
```diff
@@ -1,47 +1,54 @@
-import csv
-import io
-
-import azure.functions as func
```
bamader, line 4: Why is the git diff having a fit over this 🤷

bamader (approved):
LGTM. Honestly seems like git + the requirement to have tests is making a lot of overhead out of what amounts to a few very small changes: adding the blob configs in the __init__ function, changing some capitalization, adding a new function which just gets some creds, and modifying a signature for writing purposes. No issues here.

src/FunctionApps/python/GenerateCSVs/__init__.py
graham-skylight: `dos2unix` - azure's templates all (surprise!) have windows line endings

graham-skylight:
@bamader Yeah, this one's really not too huge, just a thing I neglected to do in the initial skeleton

PR #84, opened 2022-04-13
Read fhir for csv creation

spencer-kathol:
This PR requires updates the ingestion pipeline so that record type is preserved in the Blob Container folder structure.  So input files were already being stored as:
<input_prefix>/<record_type>/<filename>
Previously this was stored as:
<output_prefix>/<uuid>.fhir
With new changes, this is now stored as:
<output_prefix>/<record_type>/<uuid>.fhir

Where record_type is either ELR, ECR, or VXU.

The code to read fhir to be written to CSV uses the record type in the folder structure above to derive record type, and split to the appropriate record_type-specific CSV generation function.

### DIFF
--- a/src/FunctionApps/python/GenerateCSVs/fhir.py
+++ b/src/FunctionApps/python/GenerateCSVs/fhir.py
@@ -41,10 +41,12 @@ def get_blobs(
 def read_bundles_by_type(
     container_url: str, container_prefix: str, record_types: list
 ) -> Iterator[Tuple[str, dict]]:
-    """Reads FHIR bundle dicts and returns an iterator referencing tuples containing
-    the following information:
-    record_type (str) The record type (e.g. ELR, ECR, VXU)
-    bytes (BytesIO) Actual content of the container
+    """Reads FHIR bundle dicts using the specified container info
+    and returns an iterator referencing tuples containing the
+    following information:
+    record_type (str) The record type(s) to read,
+    corresponding to RECORD_TYPE constants (e.g. elr, ecr, vxu).
+    fhir_content (dict) Actual content of the container, json formatted
     """
     for record_type, fp in get_blobs(container_url, container_prefix, record_types):
         for line in fp:
--- a/src/FunctionApps/python/IntakePipeline/README.md
+++ b/src/FunctionApps/python/IntakePipeline/README.md
@@ -2,27 +2,16 @@
 
 ## Config
 
-### IngestPipeline
 This function app module requires some config to be present. In production this is defined on the function app, and locally, you'll want to add
 items in `local.settings.json` under `Values`.
 
 * `INTAKE_CONTAINER_URL`: a container URL containing FHIR bundles (eg: https://pitestdatasa.blob.core.windows.net/bronze)\
-*Note:* Stored bundles must be of type `batch`.  Entries in the batch Bundle should contain a valid `request` describing how the `resource` should be submitted to the FHIR Server.
+    * Stored bundles must be of type `batch`.  Entries in the batch Bundle should contain a valid `request` describing how the `resource` should be submitted to the FHIR Server.
 * `INTAKE_CONTAINER_PREFIX`: a string prefix so we're not scanning the entire container (eg: decrypted/valid-messages/)\
-*Note:* Assuming prefixes are a complete directory path, prefixes should have a trailing /.  Also, there must be a sub-structure of record types (all caps) under the prefix (eg: decrypted/valid-messages/VXU)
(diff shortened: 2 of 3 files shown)


src/FunctionApps/python/GenerateCSVs/fhir.py
```diff
@@ -12,17 +14,49 @@
 RECORD_TYPE_ECR = "ecr"
 
 
-def read_bundles_by_type() -> Iterator[Tuple[str, dict]]:
+def get_blobs(
```
graham-skylight, line 17: $0.02: I've been debating whether or not to make an `azure` package at the top level to deal with some of this - any thoughts there?

src/FunctionApps/python/GenerateCSVs/fhir.py
```diff
@@ -12,17 +14,49 @@
...
+                )
 
-    return
+
+def read_bundles_by_type(
+    container_url: str, container_prefix: str, record_types: list
+) -> Iterator[Tuple[str, dict]]:
+    """Reads FHIR bundle dicts and returns an iterator referencing tuples containing
+    the following information:
+    record_type (str) The record type (e.g. ELR, ECR, VXU)
```
graham-skylight, line 46: You may want to mention that the values would be defined in the constants (and I think they're lowercase)

src/FunctionApps/python/GenerateCSVs/fhir.py
```diff
@@ -12,17 +14,49 @@
...
+    bytes (BytesIO) Actual content of the container
```
graham-skylight, line 47: Isn't this yielding a `dict`?

src/FunctionApps/python/IntakePipeline/README.md
```diff
@@ -2,13 +2,27 @@
...
-This function app requires some config to be present. In production this is defined on the function app, and locally, you'll want to add
+### IngestPipeline
+This function app module requires some config to be present. In production this is defined on the function app, and locally, you'll want to add
 items in `local.settings.json` under `Values`.
 
-* `INTAKE_CONTAINER_URL`: a container URL containing ndjson formatted FHIR records (eg: https://pitestdatasa.blob.core.windows.net/bronze)
-* `INTAKE_CONTAINER_PREFIX`: a string prefix so we're not scanning the entire container (eg: decrypted/valid-messages)
-* `OUTPUT_CONTAINER_PATH`: the blob container path to store processed items
+* `INTAKE_CONTAINER_URL`: a container URL containing FHIR bundles (eg: https://pitestdatasa.blob.core.windows.net/bronze)\
+*Note:* Stored bundles must be of type `batch`.  Entries in the batch Bundle should contain a valid `request` describing how the `resource` should be submitted to the FHIR Server.
```
graham-skylight, line 10: I'd almost make these notes a separate subheader dealing with paths/contents and whatnot.

graham-skylight (approved):
Looks awesome - everything works as expected

One thing I was thinking about is if we want to combine the config `OUTPUT_CONTAINER_PATH` from `IntakePipeline` with `CSV_INPUT_PREFIX` from `GenerateCSVs` - I *think* they should be the same thing, right?

src/FunctionApps/python/GenerateCSVs/fhir.py
spencer-skylight: I think it's a good idea.  A lot of the code was copied and tweaked from IntakePipeline and could be standardized and pulled into a central place.

spencer-skylight:
> One thing I was thinking about is if we want to combine the config `OUTPUT_CONTAINER_PATH` from `IntakePipeline` with `CSV_INPUT_PREFIX` from `GenerateCSVs` - I _think_ they should be the same thing, right?

Yes, I think they'll be the same.  I considered using the existing config but thought the existing name OUTPUT_CONTAINER_PATH would be confusing in this new context.  Otherwise I'm for it.  What do you think about renaming?  Your comment got me thinking it may be clearer to drop input/output from the name... something like:
* UNENCRYPTED_BLOB_PREFIX
* TRANSFORMED_BLOB_PREFIX
* CSV_BLOB_PREFIX
* ...

Then again, maybe this is a small rabbit hole we can save for later.

src/FunctionApps/python/GenerateCSVs/fhir.py
spencer-skylight: Updated.

src/FunctionApps/python/GenerateCSVs/fhir.py
spencer-skylight: Yep, updated.

src/FunctionApps/python/IntakePipeline/README.md
spencer-skylight: Updated, and separated content into IntakePipeline and GenerateCSV specific READMEs.

PR #85, opened 2022-04-13
Add ELR parsing to CSVs

bamader:
This PR introduces the functionality needed for us to generate csv rows from ELR data. Using Dan's patient extractor function, this changeset extends that by identifying COVID related observations and incorporating relevant fields into the generated rows. Tests are included for the individual function.

Also, per slack discussion, the return type of all functions should be modified to a list of lists. I've done that for the ELR function and also set up the main functions (and tests) to work with that data type.

### DIFF
--- a/src/FunctionApps/python/GenerateCSVs/elr.py
+++ b/src/FunctionApps/python/GenerateCSVs/elr.py
@@ -2,17 +2,16 @@ from typing import List, Any
 from GenerateCSVs.patient import PATIENT_COLUMNS, parse_patient_resource
 
 
-ELR_COLUMNS = PATIENT_COLUMNS + ["loinc_code", "result", "effective_date"]
+ELR_COLUMNS = PATIENT_COLUMNS + ["loincCode", "result", "effectiveDateTime"]
 
 
-def extract_covid_test(observation: dict) -> dict:
+def extract_loinc_lab(observation: dict) -> List[str]:
     """
     Given an observation FHIR profile, determine whether the profile
-    contains information related to a COVID test or COVID status.
+    contains information related to a laboratory test and result.
     If yes, collect the relevant lab code, test result, and test
     date into a list.
     """
-    print(observation)
     code = observation["code"]["coding"][0]
     if "loinc" in code["system"] and "valueCodeableConcept" in observation:
         obs_date = observation["effectiveDateTime"]
@@ -21,19 +20,26 @@ def extract_covid_test(observation: dict) -> dict:
     return []
 
 
-def elr_to_csv(bundle: dict) -> List[Any]:
+def elr_to_csv(bundle: dict) -> List[List[str]]:
     """
     Given a FHIR bundle containing a patient resource and one or more
     ELR lab observations, identify the labs related to COVID and turn
     them into patient-identified rows for use in a CSV.
     """
     rows_to_write = []
-    row_root = []
+    patient_info_list = []
     for resource in bundle["entry"]:
         if resource["resource"]["resourceType"] == "Patient":
-            row_root.extend(parse_patient_resource(resource))
+            patient_info_list.extend(parse_patient_resource(resource))
         elif resource["resource"]["resourceType"] == "Observation":
-            obs_info = extract_covid_test(resource["resource"])
+
+            # We only care about observations related to covid testing
+            obs_info = extract_loinc_lab(resource["resource"])
             if len(obs_info) > 0:
-                rows_to_write.append(row_root + obs_info)
+                rows_to_write.append(obs_info)
+
+    # Now pre-pend the patient information to each row
+    # Handles case where Patient resource comes after 1+ obs profiles
+    for i in range(len(rows_to_write)):
+        rows_to_write[i] = patient_info_list + rows_to_write[i]
     return rows_to_write
--- a/src/FunctionApps/python/GenerateCSVs/elr.py
+++ b/src/FunctionApps/python/GenerateCSVs/elr.py
@@ -1,4 +1,4 @@
-from typing import List, Any
+from typing import List
 from GenerateCSVs.patient import PATIENT_COLUMNS, parse_patient_resource
 
 
--- a/src/FunctionApps/python/GenerateCSVs/elr.py
+++ b/src/FunctionApps/python/GenerateCSVs/elr.py
@@ -1,8 +1,39 @@
 from typing import List, Any
-from GenerateCSVs.patient import PATIENT_COLUMNS
+from GenerateCSVs.patient import PATIENT_COLUMNS, parse_patient_resource
 
-ELR_COLUMNS = PATIENT_COLUMNS + []
+
+ELR_COLUMNS = PATIENT_COLUMNS + ["loinc_code", "result", "effective_date"]
+
+
+def extract_covid_test(observation: dict) -> dict:
+    """
+    Given an observation FHIR profile, determine whether the profile
+    contains information related to a COVID test or COVID status.
+    If yes, collect the relevant lab code, test result, and test
+    date into a list.
+    """
+    print(observation)
+    code = observation["code"]["coding"][0]
+    if "loinc" in code["system"] and "valueCodeableConcept" in observation:
+        obs_date = observation["effectiveDateTime"]
+        result = observation["valueCodeableConcept"]["coding"][0]["display"]
+        return [code["code"], result, obs_date]
+    return []
 
 
 def elr_to_csv(bundle: dict) -> List[Any]:
-    pass
+    """
+    Given a FHIR bundle containing a patient resource and one or more
+    ELR lab observations, identify the labs related to COVID and turn
+    them into patient-identified rows for use in a CSV.
+    """
+    rows_to_write = []
+    row_root = []
+    for resource in bundle["entry"]:
+        if resource["resource"]["resourceType"] == "Patient":
+            row_root.extend(parse_patient_resource(resource))
+        elif resource["resource"]["resourceType"] == "Observation":
+            obs_info = extract_covid_test(resource["resource"])
+            if len(obs_info) > 0:
+                rows_to_write.append(row_root + obs_info)
+    return rows_to_write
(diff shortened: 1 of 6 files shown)


bamader:
Me: how are there so many changes in this PR it was like three functions
Also me: oh right, JSON assets 😂

DanPaseltiner (changes requested):
I think this off to a good start, but I'm pretty sure we need to make the length of each list the same to ensure that values are written to the proper column in the CSV. Also, I'm not sure the the patient resource is guaranteed to appear in the bundle before observation resources.

src/FunctionApps/python/GenerateCSVs/elr.py
```diff
@@ -1,8 +1,39 @@
 from typing import List, Any
-from GenerateCSVs.patient import PATIENT_COLUMNS
+from GenerateCSVs.patient import PATIENT_COLUMNS, parse_patient_resource
 
-ELR_COLUMNS = PATIENT_COLUMNS + []
+
+ELR_COLUMNS = PATIENT_COLUMNS + ["loinc_code", "result", "effective_date"]
```
DanPaseltiner, line 5: To align with the convention of matching FHIR variable names and using camel case could we change this to ["loincCode", "result", "effectiveDateTime"]?

bamader:
Can you elaborate on where you think some rows aren't the same length as others? The `extract_covid_test` function does return lists of different lengths, but if you look at line 37, you'll see that if an empty list is returned its not considered a row we want to write to the CSV so we don't append it. The tests also actually cover this case by including a non-covid observation, which successfully does not get written out to the CSV.

src/FunctionApps/python/GenerateCSVs/elr.py
```diff
@@ -1,8 +1,39 @@
...
+    rows_to_write = []
+    row_root = []
+    for resource in bundle["entry"]:
+        if resource["resource"]["resourceType"] == "Patient":
+            row_root.extend(parse_patient_resource(resource))
+        elif resource["resource"]["resourceType"] == "Observation":
+            obs_info = extract_covid_test(resource["resource"])
+            if len(obs_info) > 0:
+                rows_to_write.append(row_root + obs_info)
+    return rows_to_write
```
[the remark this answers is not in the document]
bamader: Oh yup you're correct. And sure, that seems like a fair assumption to eliminate. I'll get rid of that.

src/FunctionApps/python/GenerateCSVs/elr.py
```diff
@@ -1,8 +1,39 @@
...
+    If yes, collect the relevant lab code, test result, and test
+    date into a list.
+    """
+    print(observation)
+    code = observation["code"]["coding"][0]
+    if "loinc" in code["system"] and "valueCodeableConcept" in observation:
+        obs_date = observation["effectiveDateTime"]
+        result = observation["valueCodeableConcept"]["coding"][0]["display"]
+        return [code["code"], result, obs_date]
+    return []
```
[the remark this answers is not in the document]
bamader: Man my types are all over the place today. You're absolutely correct about the `List[str]`. The length though is actually important to keep different, since that's how we check for a covid observation (I did some basic profiling and data exploration before I concluded that it was both necessary and sufficient for a lab observation to have a `loinc` code and the `valueCodingConcept`). See comment above about line 37.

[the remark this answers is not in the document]
bamader: You could be right about effectiveDateTime though. Which would be really concerning, if that's the case 😕 . That's the only date like field that's consistently available across all observations. I think a deep dive on one or both our parts is probably warranted 😞 .

src/FunctionApps/python/GenerateCSVs/elr.py
```diff
@@ -1,8 +1,39 @@
...
+
+
+def extract_covid_test(observation: dict) -> dict:
+    """
+    Given an observation FHIR profile, determine whether the profile
+    contains information related to a COVID test or COVID status.
+    If yes, collect the relevant lab code, test result, and test
+    date into a list.
+    """
+    print(observation)
```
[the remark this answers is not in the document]
bamader: LOL whoops 🤣

src/FunctionApps/python/tests/GenerateCSVs/test_elr.py
```diff
@@ -0,0 +1,62 @@
...
+            "coding": [
+                {
+                    "code": "123456",
+                    "display": "some sorta health thing",
+                    "system": "http://blah.org",
+                }
+            ]
+        },
+    }
+    assert extract_covid_test(non_covid_obs) == []
```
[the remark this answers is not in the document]
bamader: See above comments

src/FunctionApps/python/GenerateCSVs/elr.py
bamader: Yup sure thing!

src/FunctionApps/python/GenerateCSVs/elr.py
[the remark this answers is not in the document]
DanPaseltiner: Ahhh ooops I missed line 37 where we check the length.

src/FunctionApps/python/GenerateCSVs/elr.py
```diff
@@ -1,8 +1,39 @@
...
+    code = observation["code"]["coding"][0]
```
spencer-skylight, line 16: I know I'm just being paranoid since code.coding is a required field, but should we be doing more checking to ensure good data?  Right now, if either code or code.coding isn't valued (or malformed), I think it will throw an error that will terminate the whole processing batch.  Same general comment applies in other areas as well.

src/FunctionApps/python/GenerateCSVs/elr.py
```diff
@@ -1,8 +1,39 @@
...
+    them into patient-identified rows for use in a CSV.
+    """
+    rows_to_write = []
+    row_root = []
+    for resource in bundle["entry"]:
+        if resource["resource"]["resourceType"] == "Patient":
+            row_root.extend(parse_patient_resource(resource))
+        elif resource["resource"]["resourceType"] == "Observation":
+            obs_info = extract_covid_test(resource["resource"])
+            if len(obs_info) > 0:
```
DanPaseltiner, line 37: Not necessary but `if len(obs_info) > 0:` can be shortened to simply `if obs_info:`. It was kind of cool to learn that Python evaluates empty lists, strings, and such False.

src/FunctionApps/python/tests/GenerateCSVs/test_elr.py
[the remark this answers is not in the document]
DanPaseltiner: You are right. My mistake.

src/FunctionApps/python/GenerateCSVs/elr.py
```diff
@@ -1,8 +1,39 @@
 from typing import List, Any
-from GenerateCSVs.patient import PATIENT_COLUMNS
+from GenerateCSVs.patient import PATIENT_COLUMNS, parse_patient_resource
 
-ELR_COLUMNS = PATIENT_COLUMNS + []
+
+ELR_COLUMNS = PATIENT_COLUMNS + ["loinc_code", "result", "effective_date"]
+
+
+def extract_covid_test(observation: dict) -> dict:
```
spencer-skylight, line 8: I'm not seeing any code here to actually identify tests are covid-related.  I think that's in line with what we decided on Slack, but should we update the function name (and adjust some test cases) so it doesn't seem like we are trying to filter to only covid results?

src/FunctionApps/python/GenerateCSVs/elr.py
bamader: We decided that for right now, we don't need to be that defensive. You're right that for a prod-like long-term solution we would want to use `try/except` here and be more defensive, but these CSVs are a one-time, one-off solution that doesn't represent any sort of long-term legitimate attempt. Our philosophy has largely been to be defensive later as the data requires us rather than get too complicated right now.

src/FunctionApps/python/GenerateCSVs/elr.py
bamader: Yeah that's fair, it's really just filtering for lab results that are test/result pairs. I can update the semantics.

src/FunctionApps/python/GenerateCSVs/elr.py
spencer-skylight: :+1: sounds good to me.

spencer-skylight (approved):
Looks good!

src/FunctionApps/python/GenerateCSVs/elr.py
```diff
@@ -1,8 +1,45 @@
...
+    them into patient-identified rows for use in a CSV.
+    """
+    rows_to_write = []
+    patient_info_list = []
+    for resource in bundle["entry"]:
+        if resource["resource"]["resourceType"] == "Patient":
+            patient_info_list.extend(parse_patient_resource(resource))
+        elif resource["resource"]["resourceType"] == "Observation":
+
+            # We only care about observations related to covid testing
```
DanPaseltiner, line 36: Technically we are grabbing any test not just COVID, so this comment isn't quite right. Super nit picky...

src/FunctionApps/python/GenerateCSVs/elr.py
```diff
@@ -1,8 +1,45 @@
...
+            obs_info = extract_loinc_lab(resource["resource"])
+            if len(obs_info) > 0:
+                rows_to_write.append(obs_info)
+
+    # Now pre-pend the patient information to each row
+    # Handles case where Patient resource comes after 1+ obs profiles
+    for i in range(len(rows_to_write)):
+        rows_to_write[i] = patient_info_list + rows_to_write[i]
+    return rows_to_write
```
DanPaseltiner, line 45: This looks like a good solution.

DanPaseltiner (approved):
LGTM. Thanks for your work on this Brandon.

PR #86, opened 2022-04-14
Convert DSTP to Python and make it part of the pipeline

bryan-skylight:
# Description
This PR takes the core functionality from the DSTP Kotlin function and converts it into Python code, which could potentially be used in the IntakePipeline setup. It would require some slight modifications, including reading in HL7 instead of FHIR bundles, parsing out the individual messages, and then processing each message instead of each pre-generated FHIR bundle, but the value add is that it moves away from the blob trigger approach that currently isn't working.

# Feedback Requested
* Does this approach seem feasible?  
What are some of the tradeoffs of this approach?  
  * An obvious one is that the IntakePipeline function will take even longer to run, which isn't great given that it already times out  
* Are there other changes that would need to be made other than what's been called out in the code?

# To Do
There's still quite a bit to be done if we were to go down this path (including writing tests, figuring out how to port the AzureDefaultCredentials functionality to Python, and tweaking the IntakePipeline approach, but I wanted to get early eyes on this to ensure it's a viable approach.


JSkeets:
Has this been tested at all ? One off on your machine or otherwise ?

bryan-skylight:
> Has this been tested at all? One off on your machine or otherwise?

No, because of as of right now it wouldn't work since the functionality to get the access token hasn't been ported to Python. This PR should have been submitted as a draft as it's not intended to be considered for merging. It was just a quick port of the main DSTP functionality to Python to help frame everyone's thoughts about what this would look like if we moved the FHIR conversion into `IntakePipeline`.

bryan-skylight:
Just pushed a major update to this PR. The code now successfully runs locally via a BlobTrigger and is able to run the code through all of our transforms (convert to FHIR; standardize name, address, and phone number; geocode; and generate patient hash) and then save the FHIR bundle to the appropriate location. It also saves the message into an invalid messages location if the conversion fails, which is happening quite a bit due to malformed datetimes. 

We'll need to confirm that the transformations are actually happening, and we'll need to make sure we can run this in Azure, but I believe this gets us closer to the ideal scenario where all transformations, including the FHIR conversion, are done in a single pipeline.

src/FunctionApps/python/IntakePipeline/.gitignore
```diff
@@ -0,0 +1,48 @@
+bin
```
graham-skylight, line 1: Is this any different than the one in the root?

src/FunctionApps/python/IntakePipeline/conversion.py
```diff
@@ -0,0 +1,153 @@
...
+    root_template = message_type_map.get(message_type.lower())
+    template_collection = template_map.get(
+        input_data_type,
+        "microsofthealth/hl7v2templates:default"
+    )
+
+    if input_data_type is None or root_template is None:
+        raise Exception("Unknown file format or message type")
+
+    url = f"{get_required_config('FHIR_URL')}/$convert-data"
```
graham-skylight, line 131: $0.02: Usually I like keeping things that read from config at a higher level (ie: as an arg to `convert_message_to_fhir`)

src/FunctionApps/python/IntakePipeline/conversion.py
```diff
@@ -0,0 +1,153 @@
...
+    data = json.dumps(data)
+
+    response = requests.post(
+        url=url,
+        data=data,
+        headers={
+            "Content-Type": "application/json",
+            "Authorization": f"Bearer {access_token}"
+        }
+    )
```
graham-skylight, line 151:
```suggestion
    response = requests.post(
        url=url,
        json=data,
        headers={
            "Authorization": f"Bearer {access_token}"
        }
    )
```

src/FunctionApps/python/IntakePipeline/function.json
```diff
@@ -2,14 +2,11 @@
...
-      "type": "httpTrigger",
+      "name": "blob",
+      "type": "blobTrigger",
       "direction": "in",
-      "name": "req",
-      "methods": [
-        "get",
-        "post"
-      ]
+      "path": "bronze/decrypted/ELR/{name}",
```
graham-skylight, line 8: was this specifically running on ELRs or is this just the last thing you were testing on?

src/FunctionApps/python/IntakePipeline/getting_started.md
```diff
@@ -0,0 +1,30 @@
+## Getting Started with Azure Function
```
graham-skylight, line 1: I'd get rid of this

src/FunctionApps/python/IntakePipeline/requirements.txt
```diff
@@ -0,0 +1,3 @@
+# Do not include azure-functions-worker as it may conflict with the Azure Functions platform
+
+azure-functions
```
graham-skylight, line 3: we've got reqiurements.txt in the python function app root

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -27,20 +31,65 @@ def run_pipeline():
     )
 
     container_url = get_required_config("INTAKE_CONTAINER_URL")
-    container_prefix = get_required_config("INTAKE_CONTAINER_PREFIX")
-    output_path = get_required_config("OUTPUT_CONTAINER_PATH")
+    valid_output_path = get_required_config("VALID_OUTPUT_CONTAINER_PATH")
```
graham-skylight, line 34: This is a lot to put in the major orchestrating function - can we refactor a little?

src/FunctionApps/python/IntakePipeline/function.json
DanPaseltiner, line 8: Is this path correct? Don't we want he blob trigger to trigger whenever there is a new blob for any message type? What about point the trigger to just bronze/decrypted?

graham-skylight (dismissed):
Glad to see us integrating all this! A bunch of minor points in comments and a few more:

* First things first, let's get this running in the pitest app - we've had lots of false starts with config and it'd be nice to prove that this can work there
* This drops test coverage by a good bit (and I'm pretty sure it'll fail tests once the conflicts are figured out). Is there a plan to fix it?

src/FunctionApps/python/IntakePipeline/getting_started.md
bryan-skylight: Done.

src/FunctionApps/python/IntakePipeline/requirements.txt
bryan-skylight: Removed this instance of it.

src/FunctionApps/python/IntakePipeline/function.json
bryan-skylight: This was the one I was testing on when I pushed. I confirmed it worked with the other file types after pushing and have now pushed the proper `function.json` file, which uses `bronze/decrypted/{name}`.

src/FunctionApps/python/IntakePipeline/.gitignore
bryan-skylight: Nope. Removed this instance of it.

bryan-skylight:
> This drops test coverage by a good bit (and I'm pretty sure it'll fail tests once the conflicts are figured out). Is there a plan to fix it?

Yes, I definitely intend to write tests for this before merging. PR should be updated by EOD 4/19.

src/FunctionApps/python/IntakePipeline/conversion.py
bryan-skylight: Makes sense to me. I've moved this outside of the function and now pass `fhir_url` in.

src/FunctionApps/python/IntakePipeline/conversion.py
bryan-skylight: Good call! This is much cleaner and has now been implemented.

src/FunctionApps/python/IntakePipeline/function.json
bryan-skylight: This was an accidental remnant of running tests and pushing before changing the path back. The code has been tested on all three file types and an updated `path` has been pushed to the PR.

bamader (approved):
Not many changes to suggest here, things look mostly good. Obviously top priority is just verifying that it actually deploys and runs.

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -5,42 +5,93 @@
...
+            message_format=message_format,
+            message_type=message_type,
+            access_token=token.token,
+            fhir_url=fhir_url
+        )
+        if response.status_code == 200:
+            bundle = response.json()
+            transform_bundle(geocoder, bundle)
+            add_patient_identifier(salt, bundle)
+            store_bundle(container_url, valid_output_path, bundle)
```
spencer-skylight, line 72: This method requires the bundle type to be supplied (elr, ecr, or vxu).  These values come from constants defined in the GenerateCSV Azure function and differ (but have a 1-1 relationship with) the filetype paramater.  These should probably be standardized across the solution.  In the short term, I think we'll need to pass either one of those values to this function - probably derived from the filetype parameter.  

Code that uses type:
https://github.com/CDCgov/prime-public-health-data-infrastructure/blob/6152260fe4e6f033d0e500fdcb4e587d478aa4e6/src/FunctionApps/python/IntakePipeline/__init__.py#L33

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -5,42 +5,93 @@
...
+        message_format = "ccda"
+    else:
+        raise Exception(f"Found an unidentified message_format: {filetype}")
 
-    for bundle in read_fhir_bundles(container_url, container_prefix):
-        transform_bundle(geocoder, bundle)
-        add_patient_identifier(salt, bundle)
-        store_bundle(container_url, output_path, bundle)
-        upload_bundle_to_fhir_server(fhirserver_cred_manager, bundle)
+    # The HL7 messages that VA sent have weird unicode characters at the beginning
```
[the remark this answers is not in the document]
bryan-skylight: `clean_message` expects a string to be passed in, but the data is in byte format, so when you try to decode to UTF to get a string, which would then be passed to `clean_message`, the error has already occurred. As a result, I think the comment should stay until we figure out how to remove those bytes before trying to decode it.

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -1,46 +1,114 @@
...
+    upload_bundle_to_fhir_server(cred_manager, bundle)
+
+
+def convert_to_fhir(
+    blob: func.InputStream,
+    fhir_url: str,
+    container_url: str,
+    invalid_output_path: str,
+    cred_manager: AzureFhirserverCredentialManager,
+) -> Tuple[str, dict]:
```
spencer-skylight, line 54: Should this method be defined in conversion.py

src/FunctionApps/python/tests/IntakePipeline/pipeline_test.py
```diff
@@ -6,48 +6,38 @@
 TEST_ENV = {
     "INTAKE_CONTAINER_URL": "some-url",
     "INTAKE_CONTAINER_PREFIX": "some-prefix",
-    "OUTPUT_CONTAINER_PATH": "output/path",
```
spencer-skylight, line 9: This change requires a terraform update too I think.

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -1,46 +1,114 @@
...
 
 
-def main(req: func.HttpRequest) -> func.HttpResponse:
+def main(blob: func.InputStream) -> func.HttpResponse:
     logging.info("Triggering intake pipeline")
     try:
-        run_pipeline()
+        fhir_url = get_required_config("FHIR_URL")
+        container_url = get_required_config("INTAKE_CONTAINER_URL")
+        invalid_output_path = get_required_config("INVALID_OUTPUT_CONTAINER_PATH")
```
spencer-skylight, line 104: We should probably add a unit test for this.

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -1,46 +1,114 @@
...
+            fhir_url=fhir_url,
+        )
 
-    for datatype, bundle in read_fhir_bundles(container_url, container_prefix):
-        transform_bundle(geocoder, bundle)
-        add_patient_identifier(salt, bundle)
-        store_bundle(container_url, output_path, bundle, datatype)
-        upload_bundle_to_fhir_server(fhirserver_cred_manager, bundle)
+        if response.status_code == 200:
+            return bundle_type, response.json()
```
bryan-skylight, line 91: Are these values reversed from what's expected in line 110? It's unpacking the values expecting the bundle to be first, but we're returning it as the second value here. Also, I think we should change line 110 to use bundle_type to keep the variable names consistent.

bryan-skylight, line 91: Are these values reversed from what's expected in line 110? It's unpacking the values expecting the bundle to be first, but we're returning it as the second value here. Also, I think we should change line 110 to use bundle_type to keep the variable names consistent.

bryan-skylight (commented):
I

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -1,46 +1,114 @@
...
-        run_pipeline()
+        fhir_url = get_required_config("FHIR_URL")
+        container_url = get_required_config("INTAKE_CONTAINER_URL")
+        invalid_output_path = get_required_config("INVALID_OUTPUT_CONTAINER_PATH")
+        cred_manager = get_fhirserver_cred_manager(fhir_url)
+
+        bundles = convert_to_fhir(
+            blob, fhir_url, container_url, invalid_output_path, cred_manager
+        )
+        for bundle, datatype in bundles:
```
spencer-skylight, line 110: This looks backwards - convert_to_fhir returns (bundle_type, bundle).

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -1,46 +1,114 @@
...
-        add_patient_identifier(salt, bundle)
-        store_bundle(container_url, output_path, bundle, datatype)
-        upload_bundle_to_fhir_server(fhirserver_cred_manager, bundle)
+        if response.status_code == 200:
+            return bundle_type, response.json()
+        else:
+            ftype, fname = blob.name.split("/")[-2:]
+            fname, ext = fname.rsplit(".", 1)
+            filename = f"{ftype.lower()}-{fname}-{i}.{ext}"
+            store_message(container_url, invalid_output_path, filename, message)
```
bryan-skylight, line 96: Not returning anything here and not raising an exception results in the code continuing to run on line 110, which results an unhandled exception where too many values are trying to be unpacked. Probably easiest to just return an empty tuple.

bryan-skylight: Even correcting for this locally is still resulting in the unpacking error. I'm looking into why this is the case and will update once I know more.

src/FunctionApps/python/IntakePipeline/__init__.py
bamader: Yeah we definitely can't have a `return` here with a singular value. Based on line 107 where we iterate over a sequence of bundles, it could make sense to create a list of tuples of `(str, dict)`, i.e. `valid_processed_bundles: List[Tuple[str, dict]]. Then when we process an individual bundle, we just access the ith tuple.

src/FunctionApps/python/IntakePipeline/__init__.py
bamader: I would agree with the empty tuple return, and then we just define a check on the length of the tuple later on when processing bundles. Or, potentially, if we take the approach I outlined of building up a list of tuples, then we may not need to return anything at all here, since we'll already have a list of the fully valid and processed bundles.

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -1,46 +1,114 @@
...
+    # VA sends \\u000b & \\u001c in real data, ignore for now
+    messages = convert_batch_messages_to_list(
+        blob.read().decode("utf-8", errors="ignore")
     )
 
-    container_url = get_required_config("INTAKE_CONTAINER_URL")
-    container_prefix = get_required_config("INTAKE_CONTAINER_PREFIX")
-    output_path = get_required_config("OUTPUT_CONTAINER_PATH")
+    for i, message in enumerate(messages):
+        token = cred_manager.get_access_token()
```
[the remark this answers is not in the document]
spencer-skylight: FWIW, the cred manager keeps track of an access token + expiration date.  This call will check the token (which is tracked in memory) isn't about to expire, and only get a new one if it needs to.  So the cost is relatively low to keeping it in the loop.  But Graham has also rightly suggested that it isn't necessary since the function only lives for bundle at a time the token really should never expire.  Note the downside of this is we get a new token for every bundle (or at least every file).  No strong preference if you change this - I think it will work either way.  Just adding some context on how it was designed.

src/FunctionApps/python/IntakePipeline/fhir.py
```diff
@@ -67,6 +67,13 @@ def store_bundle(container_url: str, prefix: str, bundle: dict, datatype: str) -
     blob.upload_blob(json.dumps(bundle).encode("utf-8"))
 
 
+def store_message(container_url: str, prefix: str, filename: str, message: str) -> None:
```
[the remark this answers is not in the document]
bryan-skylight: Great call out. I opted to rework the code a bit and generalize the `store_bundle` function to handle both valid and invalid messages. Hopefully this addresses the concern.

bryan-skylight:
I just pushed another big update to this PR. Here's a gist of what I changed:

* Refactored `main` and `run_pipeline` such that main is now responsible for iterating over the messages while run_pipeline will only handle one message at a time.  
  * This removes the `convert_to_fhir` function that was added in 9dab3c3, but still maintains the heart of the approach by keeping `run_pipeline` a 1:1 method while `main` handles the 1:N logic.  
  * The objective here is to keep the FHIR conversion inside of `run_pipeline` as I think that more closely represents our architecture diagram and better represents the idea of taking in the original data and processing it in its entirety from conversion to storage.   
* Removed all of the mappings for input data type, root template, and template collection from the `convert_message_to_fhir` function and placed them in a function that also includes the `if` tree logic that was being used to determine `bundle_type`.  
  * All of this logic now resides in a function that generates a dictionary for each file (not message) that is processed.  
* `store_bundle` has been renamed to `store_data` and now handles storing both valid and invalid messages.  
  * All files are now stored with a file name that refers to their parent file as well as a message index that should help us map any given valid or invalid file back to where it came from.  
  * This also removes a dependency on the `uuid` library.  
* `convert_message_to_fhir` now includes the logic for determining if the conversion failed or not, and returns either an empty dict (if the conversion failed) or the FHIR data. `run_pipeline` then simply checks if the dictionary is empty and proceeds as appropriate.

I believe these changes incorporate all of the feedback that was provided, both in the form of comments and code that was submitted.

I've tested this locally on my machine and confirmed that it works (at least insofar as it produces both valid and invalid messages and they're stored where/how they're supposed to be), and will be pushing another update tomorrow (04/20) that adds any necessary tests.

Feedback requested:
* Are there any changes that were requested that I missed?  
* Does this refactor achieve what was requested in the comments/code about refactoring the large `run_pipeline` code?  
* Is there any reason that we can't pass the access token and FHIR URL to `upload_fhir_bundle_to_server` to avoid needing to pass `cred_manager` both to `run_pipeline` and then to that function?

src/FunctionApps/python/tests/IntakePipeline/pipeline_test.py
```diff
@@ -6,48 +6,64 @@
...
-@mock.patch("IntakePipeline.read_fhir_bundles")
 @mock.patch("IntakePipeline.transform_bundle")
 @mock.patch("IntakePipeline.add_patient_identifier")
 @mock.patch("IntakePipeline.upload_bundle_to_fhir_server")
-@mock.patch("IntakePipeline.store_bundle")
-@mock.patch("IntakePipeline.get_fhirserver_cred_manager")
+@mock.patch("IntakePipeline.store_data")
 @mock.patch("IntakePipeline.get_smartystreets_client")
+@mock.patch("IntakePipeline.convert_message_to_fhir")
 @mock.patch.dict("os.environ", TEST_ENV)
```
graham-skylight, line 32: this is getting to be a *lot*

src/FunctionApps/python/tests/IntakePipeline/pipeline_test.py
```diff
@@ -6,48 +6,64 @@
...
+    cred_manager = mock.Mock()
+    run_pipeline(
+        "MSH|Hello World", MESSAGE_MAPPINGS, "some-fhir-url", "some-token", cred_manager
+    )
 
     patched_get_geocoder.assert_called_with("smarty-auth-id", "smarty-auth-token")
-    patched_fhir_read.assert_called_with("some-url", "some-prefix")
+    patched_converter.assert_called_with(
+        message="MSH|Hello World",
+        input_data_type=MESSAGE_MAPPINGS["input_data_type"],
```
graham-skylight, line 54: `MESSAGE_MAPPINGS` isn't being used elsewhere, right? Let's inline it (no need to build out the dict, it's a test)

graham-skylight (commented):
Glad to hear we got it working locally!

This would normally be a -1, mostly because the new code (the DSTP port, not the glue code to run the pipeline) isn't covered, and I think we could stand to simplify some parts.

That said, do we just want to merge this to unblock people and fix it after the fact? We've already got 50+ comments on this.

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -1,46 +1,100 @@
...
+        access_token = cred_manager.get_access_token()
+
+        # VA sends \\u000b & \\u001c in real data, ignore for now
+        messages = convert_batch_messages_to_list(
+            blob.read().decode("utf-8", errors="ignore")
+        )
+        message_mappings = get_file_type_mappings(blob.name)
+
+        for i, message in enumerate(messages):
+            message_mappings["filename"] = generate_filename(blob.name, i)
```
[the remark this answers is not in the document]
bryan-skylight: The way I thought about to solve this would be to move `get_file_type_mappings` into the `for` loop and generate the mappings for each message. This would be necessary because we need the message index so we know which message it maps back to in the parent file. Generating the mappings for each message felt like overkill, so this is a small price to pay to reduce the number of times the function gets called.

src/FunctionApps/python/IntakePipeline/fhir.py
```diff
@@ -21,50 +17,30 @@ def get_blob_client(container_url: str) -> ContainerClient:
...
-    bytes (BytesIO) Actual content of the container
-    """
-    for datatype, fp in get_blobs(container_url, container_prefix):
-        for line in fp:
-            try:
-                yield datatype, json.loads(line)
-            except Exception:
-                logging.exception("failed to read json contents in line, skipping file")
-                break
+def generate_filename(blob_name: str, message_index: int) -> str:
```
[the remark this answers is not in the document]
bryan-skylight: That's correct. Happy to change this to something that makes more sense.

src/FunctionApps/python/IntakePipeline/fhir.py
```diff
@@ -21,50 +17,30 @@ def get_blob_client(container_url: str) -> ContainerClient:
...
+    fname = blob_name.split("/")[-1]
+    fname, _ = fname.rsplit(".", 1)
+    return f"{fname}-{message_index}"
 
 
-def store_bundle(container_url: str, prefix: str, bundle: dict, datatype: str) -> None:
-    """Store the given bundle in the output container, in FHIR format"""
+def store_data(
```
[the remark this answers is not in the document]
bryan-skylight: Agreed, but do we really want two functions that do almost exactly the same thing?

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -1,46 +1,100 @@
...
+            run_pipeline(
+                message, message_mappings, fhir_url, access_token.token, cred_manager
```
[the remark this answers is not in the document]
spencer-skylight: I think it's fine to pass in just the token.

src/FunctionApps/python/IntakePipeline/__init__.py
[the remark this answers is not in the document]
spencer-skylight: I see both points, and probably too small a difference to worry much about.  Perhaps a middle ground would be to split the parameter into `file_type_mappings` and `message_mappings`.  The former are things that could be derived from the file, and the latter could be message-specific info (like output filename).

src/FunctionApps/python/IntakePipeline/fhir.py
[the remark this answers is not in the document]
spencer-skylight: A matter of framing I suppose.  My $0.02 is having a function that can store a blob without caring whether it's a message or bundle makes more sense.

src/FunctionApps/python/tests/IntakePipeline/pipeline_test.py
spencer-skylight: I think you're both right.  I feel like having a higher level orchestration component (eg. data factory) might be a better architectural direction.  This is OK for the pilot, but the more blocks we add to this, the more unwieldy it becomes to manage, and further away from intended use of a azure function.  We should foster this kind of thinking to apply post-pilot.

graham-skylight (approved):
:+1: tests look good

src/FunctionApps/python/tests/IntakePipeline/conversion_test.py
```diff
@@ -0,0 +1,171 @@
...
+    list3 = convert_batch_messages_to_list(TEST_STRING3)
+
+    assert len(list1) == 2
+    assert len(list2) == 0
+    assert len(list3) == 1
+
+    assert list1[0].startswith("MSH|")
+    assert list1[1].startswith("MSH|")
+    assert list3[0].startswith("MSH|")
+
```
bamader, line 71: You could _potentially_ add another set of assertions for the entries in list 1 having all three group strings in them (i.e. MSH, PID, OBX each), but that's probably overkill and these tests would satisfy me as is.

src/FunctionApps/python/tests/IntakePipeline/pipeline_test.py
bamader: @spencer-skylight Actually the goal for us is to provide a lot of different, plug and play independent blocks that STLTs can pick from at their leisure. It's actually a strength of our service that we're making so many things independent because it helps us deliver mission value, even if that does mean the organization gets kind of wonky.

bamader (approved):

PR #87, opened 2022-04-14
Add VXU Parsing for CSVs

spencer-skylight:
Add VXU Parsing for CSVs

### DIFF
--- a/src/FunctionApps/python/tests/GenerateCSVs/test_vxu.py
+++ b/src/FunctionApps/python/tests/GenerateCSVs/test_vxu.py
@@ -1,0 +1,55 @@
+import json
+import pathlib
+from GenerateCSVs.vxu import vxu_to_csv
+import pytest
+
+
+@pytest.fixture()
+def bundle():
+    return json.load(open(pathlib.Path(__file__).parent / "assets" / "vxu.json"))
+
+
+def test_elr_to_csv(bundle):
+    generated_rows = vxu_to_csv(bundle)
+    num_rows = 0
+    for row in generated_rows:
+        num_rows += 1
+        if num_rows == 1:
+            assert row == [
+                "",
+                "JOHN",
+                "DOE",
+                "1970-01-01",
+                "male",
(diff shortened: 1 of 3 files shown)


src/FunctionApps/python/tests/GenerateCSVs/test_vxu.py
```diff
@@ -0,0 +1,55 @@
...
+from GenerateCSVs.vxu import vxu_to_csv
+import pytest
+
+
+@pytest.fixture()
+def bundle():
+    return json.load(open(pathlib.Path(__file__).parent / "assets" / "vxu.json"))
+
+
+def test_elr_to_csv(bundle):
```
graham-skylight, line 12:
```suggestion
def test_vxu_to_csv(bundle):
```

src/FunctionApps/python/tests/GenerateCSVs/test_vxu.py
```diff
@@ -0,0 +1,55 @@
...
+    generated_rows = vxu_to_csv(bundle)
+    num_rows = 0
+    for row in generated_rows:
```
graham-skylight, line 15: $0.02: I liked the thing you (I think?) did the other day where you converted into a set to do unordered list equality

graham-skylight (approved):
LGTM - just the test naming

src/FunctionApps/python/tests/GenerateCSVs/test_vxu.py
spencer-skylight: fixed

src/FunctionApps/python/tests/GenerateCSVs/test_vxu.py
spencer-skylight: I simplified by simply comparing to the full expected output.  I *think* list and set equality would both run in O(n), and the expected output ordering should be deterministic so I left it as an ordered list, but I might be missing something.

PR #88, opened 2022-04-15
Update tf with recent config vals

graham-skylight:


operations/app/terraform/modules/function_app/apps.tf
```diff
@@ -40,5 +40,15 @@ module "pdi_function_app" {
     "${each.value.mi_accountName}"        = each.value.mi_accountValue
     fhir_url                              = each.value.fhir_url
     "AzureWebJobs.convertToFhir.Disabled" = each.value.AzureWebJobs_convertToFhir_Disabled
+    SMARTYSTREETS_AUTH_ID                 = "@Microsoft.KeyVault(SecretUri=https://${var.resource_prefix}-app-kv.vault.azure.net/secrets/SmartyStreetsAuthID)"
+    SMARTYSTREETS_AUTH_TOKEN              = "@Microsoft.KeyVault(SecretUri=https://${var.resource_prefix}-app-kv.vault.azure.net/secrets/SmartyStreetsAuthToken)"
```
spencer-skylight, line 44: These are currently configured in Azure as the literal values.  Have these values been loaded in to the key vault?  If so, I think Azure should be updated, and this tf representation can stay as-is.

operations/app/terraform/modules/function_app/apps.tf
```diff
@@ -40,5 +40,15 @@ module "pdi_function_app" {
...
+    CONTAINER_URL                         = "https://${var.resource_prefix}datasa.blob.core.windows.net/bronze"
+    INTAKE_CONTAINER_URL                  = "https://${var.resource_prefix}datasa.blob.core.windows.net/bronze"
+    INTAKE_CONTAINER_PREFIX               = "decrypted/valid-messages"
+    OUTPUT_CONTAINER_PATH                 = "processed"
+    CSV_INPUT_PREFIX                      = "processed"
+    CSV_OUTPUT_PREFIX                     = "csvs"
+    HASH_SALT                             = "@Microsoft.KeyVault(SecretUri=https://${var.resource_prefix}-app-kv.vault.azure.net/secrets/salt)"
```
spencer-skylight, line 51: This is also a literal value in azure.

operations/app/terraform/modules/function_app/apps.tf
```diff
@@ -40,5 +40,15 @@ module "pdi_function_app" {
     "${each.value.mi_accountName}"        = each.value.mi_accountValue
     fhir_url                              = each.value.fhir_url
     "AzureWebJobs.convertToFhir.Disabled" = each.value.AzureWebJobs_convertToFhir_Disabled
+    SMARTYSTREETS_AUTH_ID                 = "@Microsoft.KeyVault(SecretUri=https://${var.resource_prefix}-app-kv.vault.azure.net/secrets/SmartyStreetsAuthID)"
+    SMARTYSTREETS_AUTH_TOKEN              = "@Microsoft.KeyVault(SecretUri=https://${var.resource_prefix}-app-kv.vault.azure.net/secrets/SmartyStreetsAuthToken)"
+    CONTAINER_URL                         = "https://${var.resource_prefix}datasa.blob.core.windows.net/bronze"
```
spencer-skylight, line 45: Looks like we hadn't added this to Azure yet - just took care of that.

operations/app/terraform/modules/function_app/apps.tf
```diff
@@ -40,5 +40,15 @@ module "pdi_function_app" {
...
+    INTAKE_CONTAINER_URL                  = "https://${var.resource_prefix}datasa.blob.core.windows.net/bronze"
+    INTAKE_CONTAINER_PREFIX               = "decrypted/valid-messages"
```
spencer-skylight, line 47: This should have a trailing slash (it's correct in Azure and I think our code currently requires the trailing slash.

operations/app/terraform/modules/function_app/apps.tf
```diff
@@ -40,5 +40,15 @@ module "pdi_function_app" {
...
+    OUTPUT_CONTAINER_PATH                 = "processed"
+    CSV_INPUT_PREFIX                      = "processed"
+    CSV_OUTPUT_PREFIX                     = "csvs"
```
spencer-skylight, line 50: These currently differ or are missing, but probably these values are better.  Should we update Azure? Or is that going to mess with our testing?

operations/app/terraform/modules/function_app/apps.tf
```diff
@@ -40,5 +40,15 @@ module "pdi_function_app" {
     "${each.value.mi_accountName}"        = each.value.mi_accountValue
     fhir_url                              = each.value.fhir_url
     "AzureWebJobs.convertToFhir.Disabled" = each.value.AzureWebJobs_convertToFhir_Disabled
```
spencer-skylight, line 42: I see the value this references changed.  I'm not sure where to find it in the function app...

operations/app/terraform/modules/function_app/apps.tf
bryan-skylight, line 47: Should we change `INTAKE_CONTAINER_PREFIX` given the changes we're making to `IntakePipeline`? Or do we want to hold off on that until we can confirm that it runs in the cloud and make subsequent changes then?

operations/app/terraform/modules/function_app/apps.tf
```diff
@@ -40,5 +40,15 @@ module "pdi_function_app" {
...
+    SMARTYSTREETS_AUTH_ID                 = "@Microsoft.KeyVault(SecretUri=https://${var.resource_prefix}-app-kv.vault.azure.net/secrets/SmartyStreetsAuthID)"
+    SMARTYSTREETS_AUTH_TOKEN              = "@Microsoft.KeyVault(SecretUri=https://${var.resource_prefix}-app-kv.vault.azure.net/secrets/SmartyStreetsAuthToken)"
+    CONTAINER_URL                         = "https://${var.resource_prefix}datasa.blob.core.windows.net/bronze"
+    INTAKE_CONTAINER_URL                  = "https://${var.resource_prefix}datasa.blob.core.windows.net/bronze"
+    INTAKE_CONTAINER_PREFIX               = "decrypted/valid-messages"
+    OUTPUT_CONTAINER_PATH                 = "processed"
```
bryan-skylight, line 48: Same comment as above. Given the changes we're making, is this the appropriate key and/or value?

bryan-skylight (approved):
I'm not familiar with TF, but the configs seem pretty straightforward. Assuming `var.resource_prefix` works (which I've seen in other configs that Josiah put together, so I believe it does), then this looks good to me. I'm going to approve, but ideally we'd get a set of DevOps eyes on it.

operations/app/terraform/modules/function_app/apps.tf
graham-skylight: that's the idea - I added entries to keyvault for both envs

operations/app/terraform/modules/function_app/apps.tf
graham-skylight: it's not merged yet - we can always add that to the PR for when that goes in

PR #89, opened 2022-04-18
ECR to CSV parsing

spencer-skylight:
Add support for parsing ECR FHIR to CSV.


src/FunctionApps/python/GenerateCSVs/ecr.py
```diff
@@ -1,8 +1,96 @@
...
+    # Loop through bundle, and build:
+    #  * A dictionary to index patient resources
+    #  * A list of immunization records that should appear in the CSV
+    #  * A list of observation records that should appear in the CSV
+    for entry in bundle.get("entry"):
+        resource = entry["resource"]
+        if resource.get("resourceType") == "Patient" and len(resource.get("id")) > 0:
+            patient_resources["Patient/" + resource.get("id")] = parse_patient_resource(
+                entry
+            )
```
bamader, line 43: I might be wrong, but doesn't Dan's `parse_patient_resource` function already have safety checks around the size of the resource's `id` array?

src/FunctionApps/python/GenerateCSVs/ecr.py
```diff
@@ -1,8 +1,96 @@
...
+            and type(rsc.get("patient")) == dict
+            and type(rsc.get("patient").get("reference")) == str
+        ):
+            pat_ref = rsc.get("patient").get("reference")
+        elif (
+            rsc_typ == "Observation"
+            and type(rsc.get("subject")) == dict
+            and type(rsc.get("subject").get("reference")) == str
+        ):
+            pat_ref = rsc.get("subject").get("reference")
```
bamader, line 67: Nittiest of possible nits: I typically prefer fully spelled out variable names rather than too many abbreviations (i.e. patient_reference rather than pat_ref). I just find it helps smooth out readability just a bit. Not a blocker at all and feel free to ignore if you don't find it matters.

src/FunctionApps/python/GenerateCSVs/ecr.py
```diff
@@ -1,8 +1,96 @@
...
+
+        # Make sure patient reference exists, and is found in patient_resources
+        # Build a row in the CSV.
+        try:
+            if pat_ref and patient_resources.get(pat_ref):
```
bamader, line 72: This is probably overly defensive for May 3, given that we're not going to augment this beyond the spreadsheet generation of the demo, but given that it's here we don't have to pull it out

bamader (approved):
Looks good to me. Seems to be an appropriate combination of VXU and ELR data. Let's wait for a bit to see if we can get one more review on the defensiveness and the test coverage, but otherwise, I think this should be good to merge by EOD modulo a couple small changes.

src/FunctionApps/python/GenerateCSVs/ecr.py
bryan-skylight: +1 

Agree that this is non-blocking because it's a prototype that we're scrapping, but in general I would consider this a blocking change as I agree with Brandon that it affects readability.

src/FunctionApps/python/GenerateCSVs/ecr.py
```diff
@@ -1,8 +1,96 @@
...
+                if rsc_typ == "Immunization":
+                    imm_values = parse_immunization_resource(rsc)
+                    if (
+                        imm_values[0] or imm_values[1]
```
bryan-skylight, line 76: For readability, I'd prefer explicit conditions here instead of relying on Python's "truthy" and "falsey" interpretations.

src/FunctionApps/python/GenerateCSVs/ecr.py
```diff
@@ -1,8 +1,96 @@
...
+            patient_resources["Patient/" + resource.get("id")] = parse_patient_resource(
+                entry
+            )
+        elif resource.get("resourceType") in ["Immunization", "Observation"]:
+            entries.append(entry)
+
+    # Loop through immunization resources collected from bundle.
+    # For each immunization resource, pull in its corresponding patient
+    # and join immunization-specific fields.
+    for entry in entries:
```
bryan-skylight, line 50: Is there a reason to store the immunization and observation entries and then loop through them again instead of processing them in the original loop? This approach results in looping through the results twice, which I believe can be avoided if you were to create three functions that processed the entry in the initial pass. Use `process_patient_entry`, `process_immunization_entry`, and `process_observation_entry` to pass any given valid entry, determined by an initial check of `resource.get("resourceType")` and append the results to the appropriate collections.

src/FunctionApps/python/GenerateCSVs/vxu.py
```diff
@@ -48,9 +48,11 @@ def vxu_to_csv(bundle: dict) -> List[List[str]]:
         # Make sure patient reference exists, and is found in patient_resources
         # If so, build a row in the CSV.
         if pat_ref and patient_resources.get(pat_ref):
-            return_rows.append(
-                patient_resources.get(pat_ref) + parse_immunization_resource(imm_rsc)
-            )
+            imm_resource = parse_immunization_resource(imm_rsc)
+
+            # Require vaccine code or decription to build a row
+            if imm_resource[0] or imm_resource[1]:
```
bryan-skylight, line 54: I'd prefer to see explicit conditions here for readability purposes.

bryan-skylight (changes requested):

bryan-skylight:
For clarity, the blocking changes are explicitly stating the conditions in the `if` statements that reference `imm_resource[0] or imm_resource[1]`. 

The biggest change would be refactoring the code to not require looping through the entries twice, but that falls under "make it fast", which is the lowest priority in the "make it work. make it right. make it fast" mantra.

src/FunctionApps/python/GenerateCSVs/ecr.py
```diff
@@ -1,8 +1,96 @@
...
+        "immunizationVaccineCode",
+        "immunizationVaccineDescription",
+        "immunizationOccurrenceDateTime",
+    ]
+    + [  # Observation columns
+        "observationLoincCode",
+        "observationResult",
+        "observationEffectiveDateTime",
+    ]
+)
```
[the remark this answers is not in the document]
spencer-skylight: I was worried because the VXU/ELR columns are not specific enough for the broader scope of the ECR data set (e.g. loincCode).  There are a few other implications, but they are kind of a wash given this work is just for our demo/pilot.

@bamader what would you think about changing the column names in elr.py (and vxu.py) to be more specific and then implementing Dan's suggestion?

src/FunctionApps/python/GenerateCSVs/ecr.py
spencer-skylight: Sounds good - will update shortly.

src/FunctionApps/python/GenerateCSVs/ecr.py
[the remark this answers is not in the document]
bamader: I'll be honest I'm not seeing the issue with the VXU/ELR columns. I do think we should define the columns in the respective files and just import here, but Spencer, can you clarify where you think the lack of specificity is happening? If we need to change a couple column names it's not really a big deal, but since this is just a prototype one-off we should be sure the change does at least carry some value (for either us or readability purposes).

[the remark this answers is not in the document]
spencer-skylight: Maybe I'm overthinking... My concern was, for example, there is a column called "loincCode" in the ELR CSV.  Loinc codes are used pretty generally in a bunch of different places, so in an ECR you might have many different ways a loinc code could be used (order type, document type, medication route, etc).  I was suggesting that we qualify it to observationLoincCode.

But maybe we can have the best of both worlds by using list comprehension.  I just pushed a change to do this.  Let me know what you think!

bryan-skylight (approved):
LGTM

PR #91, opened 2022-04-21
Refactor FHIR bundle upload and expand test coverage

bryan-skylight:
# Description
This PR does two primary things:  
1. It modifies the `upload_bundle_to_fhir_server` function to no longer maintain its own access token using the credential manager. Instead, both the access token and the FHIR URL are passed into the function by the orchestrator. Tests were updated to reflect this.
2. Expands the coverage of the pipeline tests to also ensure the pipeline works for invalid messages. 

# Test Plan
All tests pass when running `python -m pytest` and the code has been run locally to ensure that changes to `upload_bundle_to_fhir_server` didn't cause any obvious errors. The code appears to be working at a surface level as it was before the change.

# Feedback Requested
Nothing specific. The changes mentioned in (1) are a result of feedback left in PR#86, and changes in (2) seem straightforward.

### DIFF
--- a/src/FunctionApps/python/IntakePipeline/__init__.py
+++ b/src/FunctionApps/python/IntakePipeline/__init__.py
@@ -9,7 +9,6 @@ from IntakePipeline.fhir import (
     store_data,
     generate_filename,
     get_fhirserver_cred_manager,
-    AzureFhirserverCredentialManager,
 )
 from IntakePipeline.conversion import (
     convert_batch_messages_to_list,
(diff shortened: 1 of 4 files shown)


spencer-skylight (approved):
LGTM

Just one clarification on the description:
> no longer create its own access token using the credential manager

The credential manager doesn't create a new access token on each call to get_access_token.  It maintains a local copy and only refreshes it if it is about to expire.  If the existing token's expires_on isn't within some tolerance, it simply returns.

This could be updated from "create" to "maintain".  
> no longer _obtain_ an access token using the credential manager

This may be a small point, but wanted to clarify the purpose of the credential manager is reduce network calls for a new token, while checking periodically it isn't in danger of expiring.

PR #93, opened 2022-04-21
Implement transforms metrics tracking

bamader:
This PR introduces metrics tracking into the python function app. The transform functions identify whether a patient's field has a changed value after standardization is invoked, and the result is stored as a boolean directly in the patient resource profile in accordance with FHIR extension protocols. As a result, this changeset does not actually include a script for computing metrics, rather it stores the data so that it can be written out to the appropriate CSVs and analytics computed later, where appropriate.


src/FunctionApps/python/tests/IntakePipeline/assets/patient_bundle.json
```diff
@@ -1,70 +1,71 @@
...
-          {
-            "value": "johndanger@doe.net",
-            "system": "email"
-          }
-        ]
-      }
-    }
-  ]
-}
+    "entry": [
```
graham-skylight, line 6: What actually changed here?

src/FunctionApps/python/tests/IntakePipeline/extensions_test.py
```diff
@@ -0,0 +1,87 @@
...
+        fips="36061",
+        lat=45.123,
+        lng=-70.234,
+        county_fips="dunno",
+        county_name="no idea",
+        precision="close-ish",
+    )
+    patient = find_patient_resources(bundle)[0]
+
+    expected = {
```
graham-skylight, line 36: It seems like it may be a little brittle to use the entire bundle here - what about testing just the extensions?

src/FunctionApps/python/tests/IntakePipeline/transform_test.py
```diff
@@ -86,7 +86,7 @@ def test_transform_bundle(patched_geocode, combined_bundle):
         ],
     }
 
-    transform_bundle(mock.Mock(), combined_bundle)
+    transform_bundle(mock.Mock(), combined_bundle, add_std_extension=False)
```
graham-skylight, line 89: $0.02: isn't `False` the default?

graham-skylight (changes requested):
Looks good overall, but I'm not totally convinced that it needs to be configurable, at least as is. Would it be easier to just write it out every time? Seems like it'd clean up a whole lot of the conditionals

src/FunctionApps/python/GenerateCSVs/patient.py
```diff
@@ -16,20 +18,70 @@
...
+            elif entry_type == "phone-was-standardized":
+                standardized_fields["phone"] = entry["valueBoolean"]
+            elif entry_type == "address-was-standardized":
+                standardized_fields["address"] = entry["valueBoolean"]
+        return [
+            standardized_fields["first"],
+            standardized_fields["last"],
+            standardized_fields["phone"],
+            standardized_fields["address"],
+        ]
```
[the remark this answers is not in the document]
DanPaseltiner: Or in the case when extension is not in pt_rsc return [""] * 4.

DanPaseltiner (changes requested):
Mostly nit picky things. When I wrote parse_patient_resource I tried to make sure it always returned a list of the same length so that other functions could blindly append onto it and not worry about whether values would land in the correct columns. Also, I think we probably need to update the globals defining the patient columns.

bamader:
@graham-skylight I agree with you about the conditionals and writing it out, but @DanPaseltiner and @spencer-skylight asked me to make it configurable so that we could turn off the functionality to avoid errors with the FHIR server in the future. Let's discuss either here or in a slack thread how we want to proceed with this and whether to keep them or not.

DanPaseltiner:
I initially suggested the configurability because I wasn't sure how files including the changes would play with being imported to the FHIR server. If we have done things in a way now that won't cause issues with the FHIR server I don't see a need for it.

src/FunctionApps/python/tests/IntakePipeline/assets/patient_bundle.json
bamader: I honestly have no idea. Maybe formatting or indentation?

src/FunctionApps/python/tests/IntakePipeline/extensions_test.py
bamader: Sure we can do that instead

src/FunctionApps/python/tests/IntakePipeline/transform_test.py
bamader: Luckily this won't be an issue anymore with us killing off the configurability 💪

src/FunctionApps/python/GenerateCSVs/patient.py
```diff
@@ -16,20 +18,70 @@
...
 
-def parse_patient_resource(pt_rsc: dict) -> List[str]:
-    """Given a FHIR patient resource return a list of the form:
-    [<patientHash>,<givenName>,<familyName>,<birthDate>,<gender>,<street>,<city>,
-    <state>,<postalCode>,<latitude>,<longitude>,<race>,<ethnicity>]"""
 
-    return (
+def parse_patient_resource(pt_rsc: dict, add_std_extension=False) -> List[str]:
+    """
+    Given a FHIR patient resource return whose form depends on whether or not
```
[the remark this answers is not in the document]
bamader: Yes it should, oops

src/FunctionApps/python/GenerateCSVs/patient.py
```diff
@@ -16,20 +18,70 @@
...
+        for entry in pt_rsc["extension"]:
+            entry_type = entry["url"].split("/")[-1]
+            if entry_type == "family-name-was-standardized":
+                standardized_fields["last"] = entry["valueBoolean"]
+            elif entry_type == "given-name-was-standardized":
+                standardized_fields["first"] = entry["valueBoolean"]
+            elif entry_type == "phone-was-standardized":
+                standardized_fields["phone"] = entry["valueBoolean"]
+            elif entry_type == "address-was-standardized":
+                standardized_fields["address"] = entry["valueBoolean"]
```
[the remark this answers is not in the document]
bamader: Fair enough, I can switch over to using `.get()`. It's not a hard piece of defensive coding I can implement.

src/FunctionApps/python/IntakePipeline/transform.py
```diff
@@ -13,52 +13,108 @@ def find_patient_resources(bundle: dict) -> dict:
...
-                        "url": "http://hl7.org/fhir/StructureDefinition/geolocation",
+                        "url": "http://hl7.org/fhir/StructureDefinition/geolocation",  # noqa
                         "extension": [
                             {"url": "latitude", "valueDecimal": geocoded.lat},
                             {"url": "longitude", "valueDecimal": geocoded.lng},
                         ],
                     }
                 )
+        any_dffs = any(
+            [raw_addresses[i] != std_addresses[i] for i in range(len(raw_addresses))]
```
spencer-skylight, line 113: In theory, if raw_addresses has length n, std_addresses has length n+1, and the values in the first n are the same, this could return a false positive.  Maybe consider this?
```
any_dffs = (len(raw_addresses) != len(std_addresses)) or any(
            [raw_addresses[i] != std_addresses[i] for i in range(len(raw_addresses))]
```

src/FunctionApps/python/IntakePipeline/transform.py
```diff
@@ -13,52 +13,108 @@ def find_patient_resources(bundle: dict) -> dict:
...
+        raw_phones = []
+        std_phones = []
         for telecom in patient.get("telecom", []):
             if telecom.get("system") == "phone" and "value" in telecom:
-                telecom["value"] = transform_phone(telecom["value"])
+                transformed_phone = transform_phone(telecom["value"])
+                raw_phones.append(telecom["value"])
+                std_phones.append(transformed_phone)
+                telecom["value"] = transformed_phone
+        any_diffs = any(
```
spencer-skylight, line 70: See comment about addresses.

spencer-skylight (approved):
Just one edge case to call out.  Take a look and implement if you care to.  Either way, I think it's good to go.

src/FunctionApps/python/GenerateCSVs/patient.py
```diff
@@ -14,22 +14,61 @@
...
         + [pt_rsc["resource"].get("birthDate", "")]
         + [pt_rsc["resource"].get("gender", "")]
         + get_address(pt_rsc)
         + get_race_ethnicity(pt_rsc)
+        + get_std_extensions(pt_rsc)
     )
+    return patient_resource
+
+
+def get_std_extensions(pt_rsc: dict) -> List[str]:
```
graham-skylight, line 48: Is this meaningfully tested?

src/FunctionApps/python/GenerateCSVs/patient.py
```diff
@@ -14,22 +14,61 @@
...
+            entry_type = entry["url"].split("/")[-1]
+            if entry_type == "family-name-was-standardized":
+                standardized_fields["last"] = entry.get("valueBoolean")
+            elif entry_type == "given-name-was-standardized":
+                standardized_fields["first"] = entry.get("valueBoolean")
+            elif entry_type == "phone-was-standardized":
+                standardized_fields["phone"] = entry.get("valueBoolean")
+            elif entry_type == "address-was-standardized":
+                standardized_fields["address"] = entry.get("valueBoolean")
+    return list(standardized_fields.values())
```
DanPaseltiner, line 72: This looks a lot cleaner now and preservers the length of the list. Thanks!

DanPaseltiner (approved):
LGTM

DanPaseltiner:
This code looks good to me. Before merging could we run this function app locally to process data and generate csvs before merging and potentially deploying. There might be something we are all missing. We can always roll back if needed though...

PR #95, opened 2022-04-25
Enhance error logging

spencer-skylight:
## Summary of Changes
The following changes were made in this PR: 
* .gitignore .python-version (pyenv), and different versions developers might maintain for local.settings.json connecting to different environments 
* Catch exceptions where resources are processed twice.  Rather than throw an exception that halts the entire processing batch, log a warning, and continue.
* Remove vestages of http responses that were causing errors for blob triggers
* Track file suffix (hl7, xml) so invalid-messages written to blob storage have the correct suffix
* Improve logging for failed $convert requests
* Unit tests


src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -55,26 +58,39 @@ def run_pipeline(
...
-        )
+        try:
+            store_data(
+                container_url,
+                valid_output_path,
+                f"{message_mappings['filename']}.fhir",
+                message_mappings["bundle_type"],
+                bundle=bundle,
+            )
+        except ResourceExistsError:
```
graham-skylight, line 69: Wouldn't we want to be able to stomp an existing resource if we're processing it differently?

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -55,26 +58,39 @@ def run_pipeline(
...
+                message=message,
+            )
+        except ResourceExistsError:
+            logging.warning(
+                "Attempted to store preexisting resource: "
+                + f"{message_mappings['filename']}.{message_mappings['file_suffix']}"
+            )
 
 
 def main(blob: func.InputStream) -> func.HttpResponse:
```
graham-skylight, line 92:
```suggestion
def main(blob: func.InputStream) -> None:
```

src/FunctionApps/python/IntakePipeline/conversion.py
```diff
@@ -152,6 +156,42 @@ def convert_message_to_fhir(
...
+                        error_info = single_error_info
+                    else:
+                        error_info += "\n\t" + single_error_info
+
+        except Exception:
+            # ; If an exception occurs while parsing FHIR JSON,
+            # Log the full response content
+            decoded_response = response.content.decode("utf-8")
+            error_info = f"HTTP Code: {response.status_code}, "\
+            + f"Response Content {decoded_response}"
```
graham-skylight, line 191: $0.02: `logging.exception` here may give some extra details about why it couldn't parse correctly

src/FunctionApps/python/IntakePipeline/conversion.py
```diff
@@ -152,6 +156,42 @@ def convert_message_to_fhir(
...
 
     if response.status_code != 200:
+
+        error_info = ""
+
+        # Try to parse the non-success response as OperationOutcome FHIR JSON
+        try:
+            response_json = response.json()
+
+            logging.info(str(response_json))
```
graham-skylight, line 166:
```suggestion
            logging.info("non-200 response from fhir converter: " + str(response_json))
```

graham-skylight (commented):
:+1: with a couple very minor things

src/FunctionApps/python/IntakePipeline/conversion.py
```diff
@@ -187,10 +187,12 @@ def convert_message_to_fhir(
...
             decoded_response = response.content.decode("utf-8")
-            error_info = f"HTTP Code: {response.status_code}, "\
-            + f"Response Content {decoded_response}"
+            error_info = (
+                f"HTTP Code: {response.status_code}, "
+                + f"Response Content {decoded_response}"
+            )
 
-        logging.error(f"Error during $convert-data -- {error_info}")
+        logging.exception(f"Error during $convert-data -- {error_info}")
```
graham-skylight, line 195: I could be wrong, but I think `logging.exception` is really only for `except` blocks

graham-skylight (approved):
LGTM

PR #19, opened 2022-02-02
Updating documentation directory, including info about local development environment

NatHillardUSDS:
moving existing documentation to docs folder, adding documentation around getting started with local development

Partially addresses https://app.clickup.com/t/27bb01u , we will add additional info about kotlin dev at a later time

PR #98, opened 2022-04-28
Add script to create a COVID case datamart from a set of ELR record

DanPaseltiner:
This PR adds a script to generate a covid case datamart from ELR records and writes it back Azure blob storage. There is also some minor refactoring to move functions that are shared between `analyze_missingness.py` and `generate_covid_case_datamart.py` into `utils/`.

### DIFF
--- a/src/scripts/assets/covid_lab_results_standard.json
+++ b/src/scripts/assets/covid_lab_results_standard.json
@@ -1,0 +1,18 @@
+{
+    "POSITIVE": [
+        "POSITIVE",
+        "DETECTED",
+        "REACTIVE"
+    ],
+    "NEGATIVE": [
+        "NEGATIVE",
+        "NOT DETECTED",
+        "NON-REACTIVE",
+        "(NOTE) NOT DETECTED"
+    ],
+    "INCONCLUSIVE": [
+        "INCONCLUSIVE",
+        "INVALID",
+        "INVALID RESULT"
+    ]
+}
--- a/src/scripts/generate_covid_case_datamart.py
+++ b/src/scripts/generate_covid_case_datamart.py
@@ -1,0 +1,111 @@
+import pandas as pd
+import pathlib
+import json
+from typing import List, Callable
+from utils import read_blob, record_combination_func, write_blob
+
+
+def standardize_lab_result(result: str) -> str:
+    """
+    Given a covid test result and dictionary whose keys (POSITIVE, NEGATIVE, and
+    INCONCLUSIVE) represent standard results and whose values are lists containing
+    non-standard results that map to the associated standard result then return a
+    standardized result. Return null is no standard result could not be found.
+    """
+    standard = json.loads(
+        open(
+            pathlib.Path(__file__).parent / "assets" / "covid_lab_results_standard.json"
+        ).read()
+    )
+    result = result.upper()
+    for standard_result, non_standard_results in standard.items():
+        if result in non_standard_results:
+            return standard_result
+
+    return ""
+
+
+def identify_covid_cases(
+    patient_labs: pd.DataFrame, agg_function: Callable
+) -> List[pd.Series]:
+    """
+    Given a data frame containing a single patient's COVID labs and an aggregation
+    function, review the labs for possible COVID cases and return a list containing a
+    for each identified case. Data for fields specific to the patient and not the
+    infection result from aggregating accross all labs.
+
+    **** WARNING ****
+
+    As currently written this function applies an approximation of the official CDC/CTSE
+    COIVD case defintion. In its current form this function is intended as a proof of
+    concept for development purposes.
+
+    Known Limitations:
+    1.  Suspect cases resulting from antibody tests are not considered.
+    2.  No check has been implemented for a negative PCR within 48 hours of a positive
+        Antigen that would negate the case.
+    3.  No consideration has been given to how the case definition has changed over the
+        course of the pandemic.
+    """
+    case_list = []
+    patient_labs.sort_values(by=["effectiveDateTime"], ascending=True)
+    positive_labs = patient_labs.loc[patient_labs.result == "POSITIVE"].reset_index()
+    remaining_labs = len(positive_labs)
+    while remaining_labs > 0:
+        collection_date = positive_labs.effectiveDateTime[0]
+        current_infection_labs = positive_labs.loc[
+            positive_labs.effectiveDateTime.map(lambda x: (x - collection_date).days)
+            <= 90
+        ]
+        if "pcr" in current_infection_labs.type.to_list():
+            status = "confirmed"
+        else:
+            status = "probable"
+        case = patient_labs.agg(record_combination_func)
+        case["effectiveDateTime"] = collection_date
+        case["status"] = status
+        case_list.append(case)
+        positive_labs = positive_labs.loc[
+            positive_labs.effectiveDateTime.map(lambda x: (x - collection_date).days)
+            > 90
+        ].reset_index()
+        remaining_labs = len(positive_labs)
+    return case_list
+
+
+if __name__ == "__main__":
+
+    # Set values the specify a blob to load.
+    STORAGE_ACCOUNT_URL = "https://pitestdatasa.blob.core.windows.net"
+    CONTAINER_NAME = "bronze"
+    ELR_FILE_NAME = "csv-test/elr.csv"
+    COVID_CASE_DATAMART_FILENAME = "datamart-test/covid_case_datamart.csv"
+
+    # Load data.
+    labs = pd.read_csv(read_blob(STORAGE_ACCOUNT_URL, CONTAINER_NAME, ELR_FILE_NAME))
+    covid_loincs = pd.read_csv(
+        pathlib.Path(__file__).parent / "assets" / "covid_loinc_reference.csv"
+    )
+
+    # Clean and standardize data.
+    covid_labs = pd.merge(labs, covid_loincs, how="inner", on="loincCode")
+    covid_labs = covid_labs.loc[covid_labs.type.isin(["pcr", "ag"])]
+    covid_labs.drop(["loincCode", "notes"], axis=1, inplace=True)
+    covid_labs.result = covid_labs.result.map(standardize_lab_result)
+    covid_labs.effectiveDateTime = pd.to_datetime(covid_labs.effectiveDateTime)
+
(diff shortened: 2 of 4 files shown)


src/scripts/assets/covid_lab_results_standard.json
```diff
@@ -0,0 +1,18 @@
...
+        "NOT DETECTED",
+        "NON-REACTIVE",
+        "(NOTE) NOT DETECTED"
+    ],
+    "INCONCLUSIVE": [
+        "INCONCLUSIVE",
+        "INVALID",
+        "INVALID RESULT"
+    ]
+}
```
JSkeets, line 18: nitpick: newline

src/scripts/utils/__init__.py
```diff
@@ -0,0 +1,54 @@
+from azure.identity import DefaultAzureCredential
+from azure.storage.blob import BlobClient
+from io import StringIO
+import pandas as pd
+
+
+def get_blob_client(url: str, container: str, file: str) -> BlobClient:
```
JSkeets, line 7: This is a nice refactor to extract these out to this utils folder

src/scripts/generate_covid_case_datamart.py
```diff
@@ -0,0 +1,111 @@
...
+        positive_labs = positive_labs.loc[
+            positive_labs.effectiveDateTime.map(lambda x: (x - collection_date).days)
+            > 90
+        ].reset_index()
+        remaining_labs = len(positive_labs)
+    return case_list
+
+
+if __name__ == "__main__":
+
```
JSkeets, line 77: This is just food for thought, should these variables be read from some sort of env variable file ? Is this the best place for these ?

JSkeets (approved):
Looks good to me, nice refactor, great code comments.

src/scripts/generate_covid_case_datamart.py
DanPaseltiner: Yeah, I have been wondering the same thing. I don't have a sense of how this code might live on beyond 5/3 or what it looks like "in production" beyond just being simple .py files to run locally ad-hoc. If you have thoughts I'm open to them. Happy to refactor into a config file for now if you think it is worth it.

src/scripts/assets/covid_lab_results_standard.json
DanPaseltiner: I'm not sure I follow you here. You would like me to add newline (#19)?

src/scripts/assets/covid_lab_results_standard.json
```diff
@@ -0,0 +1,18 @@
+{
+    "POSITIVE": [
+        "POSITIVE",
+        "DETECTED",
+        "REACTIVE"
```
bamader, line 5: Dan, just so I understand, these are the state derived from the loinc codes that _definitively_ capture a patient's infection status? Do we know if this list is exhaustive and comprehensive, or are there other resources we need to check to add potential statuses?

src/scripts/generate_covid_case_datamart.py
```diff
@@ -0,0 +1,131 @@
...
+from typing import List, Callable
+from utils import read_blob, record_combination_func, write_blob
+
+
+def standardize_lab_result(result: str) -> str:
+    """
+    Given a covid test result and dictionary whose keys (POSITIVE, NEGATIVE, and
+    INCONCLUSIVE) represent standard results and whose values are lists containing
+    non-standard results that map to the associated standard result then return a
+    standardized result. Return null is no standard result could not be found.
```
bamader, line 13: smallest possible nit: you mean `if no` but have `is no`, I'll see myself out 😂

src/scripts/generate_covid_case_datamart.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+    standard = json.loads(
+        open(
+            pathlib.Path(__file__).parent / "assets" / "covid_lab_results_standard.json"
+        ).read()
+    )
```
bamader, line 19: We could probably load the standards dict outside of this function, which seems to handle mapping a singular result (or small set of results). I would imagine the most appropriate place for this is right before a big iterator loop over a set of files / results. IO can be fast, but because the dictionary is so small I see no disadvantage to just keeping it in memory.

src/scripts/generate_covid_case_datamart.py
```diff
@@ -0,0 +1,131 @@
...
+    result = result.upper()
+    for standard_result, non_standard_results in standard.items():
+        if result in non_standard_results:
+            return standard_result
```
bamader, line 23: We could also make this code faster and more efficient by inverting the dictionary before we process the result. If we have a dict of the form `non-standard-result --> standardized result` then we don't have to loop over the iterators inside the dictionary for every result we process. Inverting a dictionary is also super easy, so in conjunction with my comment above about moving the dict loading to outside of whatever loop we do, let's just add some code to invert, e.g.

```
nonstandard_to_standard_mappings = {}
for std_res in standard.keys():
    for nonstd_res in standard[std_res]:
        nonstandard_to_standard_mappings[nonstd_res] = std_res
```

src/scripts/generate_covid_case_datamart.py
```diff
@@ -0,0 +1,131 @@
...
+        else:
+            status = "probable"
+        case = patient_labs.agg(record_combination_func)
+        case["effectiveDateTime"] = collection_date
+        case["status"] = status
+        case_list.append(case)
+        positive_labs = positive_labs.loc[
+            positive_labs.effectiveDateTime.map(lambda x: (x - collection_date).days)
+            > 90
+        ].reset_index()
```
bamader, line 91: Total non-blocker, but have we investigated where there's a more "pandas" way to do this? The functionality in here is really custom / hand-spun so there might not be, but when using pandas I know my first instinct is always to sort values and use loops when pandas can handle a lot of that for me, so I have to go back and make those changes. Just wanted to check, but if this is the best way, then totally good with me.

src/scripts/generate_covid_case_datamart.py
bamader: Personally I'd just leave them in the .py file for right now. Precisely because we don't know the answers to those (good) questions, I have zero issue with our "analytics" workflow being "some dude running scripts on his laptop" ("some dude" in this case being one of us 🤣 ). If we want to productionize this for more robust analysis at a later date we'll want to make other changes anyway so imo this is fine as is.

src/scripts/generate_covid_case_datamart.py
```diff
@@ -0,0 +1,131 @@
...
+    # Load data.
+    labs = pd.read_csv(read_blob(STORAGE_ACCOUNT_URL, CONTAINER_NAME, ELR_FILE_NAME))
+    covid_loincs = pd.read_csv(
+        pathlib.Path(__file__).parent / "assets" / "covid_loinc_reference.csv"
+    )
+
+    # Clean and standardize data.
+    covid_labs = pd.merge(labs, covid_loincs, how="inner", on="loincCode")
+    covid_labs = covid_labs.loc[covid_labs.type.isin(["pcr", "ag"])]
+    covid_labs.drop(["loincCode", "notes"], axis=1, inplace=True)
```
bamader, line 113: Just a fun tip, when you're invoking `.drop()` you can actually just give it a `columns` parameter instead of having to muck around with an axis, e.g. `covid_labs.drop(columns=['loincCode', 'notes'])`.

src/scripts/generate_covid_case_datamart.py
```diff
@@ -0,0 +1,131 @@
...
+    covid_labs.result = covid_labs.result.map(standardize_lab_result)
```
bamader, line 114: We can write this `.map()` function to take advantage of the dictionary inversion and single loading I suggested above by just substituting a lambda, e.g. `covid_labs.result = covid_labs.result.map(lambda x: standardize_lab_result(x, nonstd_to_std_mappings))` where we've previously run the dictionary inversion code to define `nonstd_to_std_mappings`.

bamader (approved):
Code looks legit to me. I did suggest a few enhancement changes, but nothing on the order that I think should block this PR if others want it merged (just some small improvements). I'm good either way!

src/scripts/assets/covid_lab_results_standard.json
DanPaseltiner: There is no exhaustive list because it is whatever the labs/hospitals happen to report which is sometimes subject to equipment changes and human error.

src/scripts/assets/covid_lab_results_standard.json
bamader: (my guess is yes? Typically when github shows that minus sign with a circle it's mad that there isn't a trailing newline)

src/scripts/assets/covid_lab_results_standard.json
bamader: Mmmm got it. I would love to see us use an ML identification tool to pull things like that out, but that's a ways down the road so I'm perfectly happy with this for the time being.

src/scripts/generate_covid_case_datamart.py
bamader: 🤮  okay that makes sense. Dask would almost certainly be the way to go (or potentially Vaex, that's got big data level simultaneous vectorization--I did a bunch of work for a CMS proposal using both of those), but if you've already looked into "pandas-izing" this then I don't feel the need to belabor the point here. If we make analytics more robust down the line we can revisit a more production-ready notebook anyway.

src/scripts/generate_covid_case_datamart.py
DanPaseltiner: fixed.

src/scripts/generate_covid_case_datamart.py
DanPaseltiner: Not going to bother with inverting the dict for now, but I moved loading it and pass it using a lambda function as you suggest.

src/scripts/generate_covid_case_datamart.py
DanPaseltiner: I'm going to resolve this conversation with the understanding we will need to revisit it in the future. but seems beyond the scope of this PR.

src/scripts/generate_covid_case_datamart.py
DanPaseltiner: Moved out of this function. See comment below.

PR #99, opened 2022-04-28
Timestamp normalization

spencer-skylight:
## Summary
Previously, malformed timestamps in MSH-7 prevented some messages from processing.  Now, processing is more lenient - if additional precision is added to different parts of an otherwise valid HL7 timestamp format, the value is modified to conform to stricter format validation and allowed to process.


src/FunctionApps/python/IntakePipeline/conversion.py
```diff
@@ -1,18 +1,62 @@
...
-def clean_message(message: str, delimiter: str = "\n") -> str:
-    cleaned_message = re.sub("[\r\n]+", delimiter, message)
+def clean_message(message: str) -> str:
+    parsed_message = None
+    try:
+        parsed_message = hl7.parse(message)
+
+        # MSH-7
+        cleansed_datetime = normalize_hl7_datetime(parsed_message[0][7][0])
+        parsed_message[0][7][0] = cleansed_datetime
```
bamader, line 15: How confident are we that the datetime that was causing the issue will always occur in that position? I'm a little nervous of there being other dates buried in the messages, or even of the structure of the message being such that some other information content is in this place instead of a datetime.

src/FunctionApps/python/IntakePipeline/conversion.py
```diff
@@ -1,18 +1,62 @@
...
+
+    except Exception:
+        logging.exception(
+            "Exception occurred while cleaning message.  "
+            + "Passing through original message."
+        )
+
+        return message
```
bamader, line 23: How robust are the failure conditions of `hl7.parse()`? That seems to be a new imported library so if we're pinning our parsing to that we should understand what it can and can't do. I actually at one point wrote a notebook whose purpose is now long passed but that used Python's hl7 parsing library (should be in `src/notebooks/ELR Profiling Template`). When I coded that up, I remember the docs actually recommending using newlines to delimit messages within a batch, so if you check the code you can see that I had to build up the message list through concatenation. I think I even pulled that straight from an example off their site. Do we need to be doing something like that here?

src/FunctionApps/python/IntakePipeline/conversion.py
```diff
@@ -1,18 +1,62 @@
...
+    return str(parsed_message)
+
+
+def normalize_hl7_datetime(hl7_datetime: str) -> str:
+    hl7_datetime_parts = re.match(
+        r"(\d+)(\.)?(\d+)?([+-])?(\d+)?", hl7_datetime
+    ).groups()
+
+    # Date Base
+    normalized_datetime = hl7_datetime_parts[0][:14]  # First 14 digits
```
bamader, line 34: This also makes me nervous. Should we be converting incoming datetime strings into some kind of known format using Python functions or the `datetime` modules before we chunk the first 14 digits? I don't know why programmers still haven't figured out format standardization around dates but I feel like there's literally always some format spec error.

src/FunctionApps/python/IntakePipeline/conversion.py
```diff
@@ -1,18 +1,62 @@
...
+    if hl7_datetime_parts[2]:
+        normalized_datetime += (
+            hl7_datetime_parts[1] + hl7_datetime_parts[2][:4]
+        )  # . plus first 4 digits
+
+    # Date Timezone
+    if hl7_datetime_parts[4] and len(hl7_datetime_parts[4]) >= 4:
+        normalized_datetime += (
+            hl7_datetime_parts[3] + hl7_datetime_parts[4][:4]
+        )  # +/- plus 4 digits
```
bamader, line 46: Same comment as above. Would a built-in library be suitable to more robustly handle these formatting conversions?

src/FunctionApps/python/tests/IntakePipeline/assets/FileSingleMessageLongTZ.hl7
```diff
@@ -1,10 +1,6 @@
-FHS|^~\&|WIR11.3.2|WIR|||20200514||1219144.update|||
-BHS|^~\&|WIR11.3.2|WIR|||20200514|||||
-MSH|^~\&|WIR11.3.2^^|WIR^^||WIRPH^^|20200514||VXU^V04|2020051411020600|P^|2.4^^|||ER
-PID|||3054790^^^^SR^~^^^^PI^||ZTEST^PEDIARIX^^^^^^|HEPB^DTAP^^^^^^|2018080812345235|M|||||||||||||||||||||
```
bamader, line 4: How come we're deleting lines out of this test case? If it's supposed to be a bad test I would think we still want the parser robust against whatever these lines and the below ones were testing.

src/FunctionApps/python/tests/IntakePipeline/conversion_test.py
```diff
@@ -1,26 +1,65 @@
...
+    datetime_6 = "20200514.123456-070000"
+    datetime_7 = "20200514010000.1234-0700"
+
+    assert normalize_hl7_datetime(datetime_1) == "20200514010000"
+    assert normalize_hl7_datetime(datetime_2) == "20200514010000"
+    assert normalize_hl7_datetime(datetime_3) == "20200514"
+    assert normalize_hl7_datetime(datetime_4) == "20200514.1234"
+    assert normalize_hl7_datetime(datetime_5) == "20200514+0400"
+    assert normalize_hl7_datetime(datetime_6) == "20200514.1234-0700"
+    assert normalize_hl7_datetime(datetime_7) == "20200514010000.1234-0700"
```
bamader, line 62: Regardless of my earlier comments about formatting robustness, this is at least super encouraging to see, so good job on the test coverage here 😌

bamader (commented):
Looks like a good start, just some comments and questions around the behavior of format strings we're using as well as the HL7 library.

src/FunctionApps/python/IntakePipeline/conversion.py
```diff
@@ -1,18 +1,130 @@
...
+        
+        # SPM-18 Specimen received date/time
+        # SPM-19 Specimen expiration date/time
+        normalize_hl7_datetime_segment(parsed_message, "SPM", [18,19])
+
+        # RXA-3 Date/time start of administration
+        # RXA-4 Date/time end of administration
+        # RXA-16 Substance expiration date
+        # RXA-22 System entry date/time
+        normalize_hl7_datetime_segment(parsed_message, "RXA", [3,4,16,22])
```
DanPaseltiner, line 59: Could we condense all of the calls of normalize_hl7_datetime_segement to looping over a dictionary with key:value pairs of the form hl7_segment:list of indexes? Might make it easier to update the places we look for dates in the future. Also possibly to support adapting to other  HL7 formats. This is clean and well documented as is though.

src/FunctionApps/python/IntakePipeline/conversion.py
DanPaseltiner: I'm a little leery of the RegEx here as well, but I'm happy to be convinced it is sufficiently robust through testing. Definitely a simple solution.

DanPaseltiner (commented):
This is looks good, clean and well commented. Holding off on a +1 for now because I would like hear Spencer's thoughts on my comments below before merging.

src/FunctionApps/python/tests/IntakePipeline/assets/FileSingleMessageLongTZ.hl7
spencer-skylight: I had uploaded this batch message with another ticket, but realized there weren't actually tests written around it, so I renamed and modified it to fit the testing I wanted to do for this.

src/FunctionApps/python/IntakePipeline/conversion.py
spencer-skylight: You were absolutely right about newlines.  I was just handling MSH before so I didn't run in to any issues.  I've updated the code to handle all segments I found in our input files, and all date/time fields that are outlined in the HL7 specs.  As a result the newline issue is definitely a blocker, and I've updated the code to handle it.

src/FunctionApps/python/IntakePipeline/conversion.py
spencer-skylight: I've been doing some extensive testing and it seems to meet our needs.  It is admittedly narrowly defined but works for this use case.

DanPaseltiner: Works for me. I'm okay to move forward with this and keep an eye on things as more data moves through the pipeline. Also totally appreciate the initial intent was to get this done for demo.

src/FunctionApps/python/IntakePipeline/conversion.py
bamader: This is resolved above and beyond my concerns. Really really well done here, Spencer!

src/FunctionApps/python/IntakePipeline/conversion.py
bamader: I agree, I'm thoroughly satisfied with the amount of testing that went into this. I'm perfectly convinced that we can use these regexes here until proven otherwise.

bamader (approved):
This is quite simply an excellent pull request. Code is clean, well-documented, rationalization and decision logs are contained throughout, and all previous comments were tested intensively and addressed in as thorough of a manner as possible. Spencer, you knocked it out of the park on this one (and frankly upped the standard for what a clean and tested PR looks like lol 😅 ). Great work!

spencer-skylight:
> This is quite simply an excellent pull request...

Thank you @bamader! I'm blushing. I appreciate all your feedback.

PR #100, opened 2022-04-28
Add cli option to generate csvs

graham-skylight:


src/FunctionApps/python/GenerateCSVs/README.md
```diff
@@ -10,3 +10,19 @@ This function app requires config to be present. Locally these are settings in `
     * There must be a sub-structure of record types (all caps) under the prefix (eg: `[CSV_INPUT_PREFIX]`VXU, `[CSV_INPUT_PREFIX]`ELR)
 * `CSV_OUTPUT_PREFIX`: a string prefix where FHIR CSV output files are stored within the container
     * Do not include a trailing /.
+
+## Running via CLI
+
+Run `generate_csvs.py` from the python function root via something like:
+
+```
+python generate_csvs.py "https://pitestdatasa.blob.core.windows.net/bronze" "blob-trigger-out/valid-messages/" vxu > /tmp/vxu.csv"
```
bamader, line 19: Does this mean the script version doesn't iterate over the given directory for all three filetypes anymore?

src/FunctionApps/python/generate_csvs.py
```diff
@@ -0,0 +1,33 @@
...
+
+    writer = csv.writer(sys.stdout)
+    writer.writerow(cols)
+
+    for _, bundle in read_bundles_by_type(container_url, prefix, [record_type]):
+        writer.writerows(fn(bundle))
+
+
+if __name__ == "__main__":
+    typer.run(main)
```
bamader, line 33: In case anyone else was confused about this, `typer` is a library that basically lets the parameters to `main()` define the expected CLI parameters. That's why there's no `sys.argv` crap in here.

src/FunctionApps/python/GenerateCSVs/README.md
bamader: Or rather, it looks like all of the three data types are being combined into a single file, since there aren't three output files anymore? Not clear on how this is happening but I would definitely like this tool to be able to generate three files, since we'll want to combine them or not for depending on our use case (e.g. for May 3 we may just be looking at ECR + ELR equity stuff).

bamader (commented):
Looks good, just a question around file combination and the output files.

src/FunctionApps/python/GenerateCSVs/README.md
DanPaseltiner: Looks like this will do just one record type at a time, so we would need to run generate_csvs.py 3 times to make all 3 csvs. Given the current situation this seems fine to me.

DanPaseltiner (approved):
LGTM. Thanks for nocking this out.

bamader:
Me: makes comment about `typer` obsoleting the argv stuff because it just takes arguments as parameters.
Also me: completely misses the parameter for filetype 🙄

bamader (approved):

src/FunctionApps/python/GenerateCSVs/README.md
graham-skylight: it's one record type at a time - figured it may be helpful in case

a. we're having issues with only one type or
b. we want to try to run them in parallel

PR #103, opened 2022-05-02
Loinc codes, azure fix, analysis fix

bamader:
This PR does a couple things:
- smooths loinc code parsing to adapt to the new observation structure of ecr data
- implements a fix to Azure credential retrieval for use with `az login` (previously there were issues with shared tokens being cached which completely destroys az login ability, it's an absolutely terrible feature but this is the consensus fix)
- adds a `vax_codes` for vxu to join on to rule out non covid vaccines
- provides the updated version of `analyze_missingness.py` for equity analysis


src/FunctionApps/python/GenerateCSVs/elr.py
```diff
@@ -17,12 +17,18 @@ def extract_loinc_lab(observation: dict) -> List[str]:
...
+        if "valueCodeableConcept" in observation:
+            try:
+                result = observation["valueCodeableConcept"]["coding"][0]["display"]
+            except KeyError:
+                result = ""
+        elif "valueString" in observation:
+            try:
+                result = observation["valueString"]
+            except KeyError:
+                result = ""
```
DanPaseltiner, line 31: Do we need a:
```
else:
     result = ""
```
at the end of the if block so there is always a value of result to return?

src/FunctionApps/python/GenerateCSVs/fhir.py
```diff
@@ -59,7 +59,7 @@ def read_bundles_by_type(
 
 def get_container_client(url: str):
     """Use whatever creds Azure can find to authenticate with the storage container"""
-    creds = DefaultAzureCredential()
+    creds = DefaultAzureCredential(exclude_shared_token_cache_credential=True)
```
DanPaseltiner, line 62: What does this do?

src/scripts/analyze_missingness.py
```diff
@@ -92,16 +99,30 @@ def count_records_missing_data(
...
+    covid_vax_codes = pd.read_csv(
+        pathlib.Path(__file__).parent / "assets" / "covid_vax_codes.csv", dtype="object"
+    )
 
-    pre_linkage_covid = pre_linkage.merge(covid_loincs, on="loincCode", how="inner")
+    pre_linkage_elr = pre_linkage_elr.merge(covid_loincs, on="loincCode", how="inner")
+    pre_linkage_ecr = pre_linkage_ecr.merge(covid_loincs, on="loincCode", how="inner")
+    pre_linkage_vxu = pre_linkage_vxu.merge(
+        covid_vax_codes, on="vaccineCode", how="inner"
+    )
```
DanPaseltiner, line 113: How do we solve the case where an ECR contains multiple covid tests and similarly when a VXU contains multiple COVID immunizations?

src/FunctionApps/python/GenerateCSVs/elr.py
bamader: No, because there's a check on the length of the array created here (line 84 in `ecr.py` and line 51 in `elr.py`) that uses length to determine whether we should write the result to a row. Essentially the defensive coding bit is just moved to the caller instead of this function.

src/FunctionApps/python/GenerateCSVs/fhir.py
bamader: So this solves a problem I was having over the weekend trying to get azure to work. I tried to `az login`, which succeeded, but then when I ran the `generate_csvs.py` script locally, I got azure authentication errors. It was like a 7 hour debugging process, but the long and short of it is that somehow, the credential I used to log in as a superuser had been cached somewhere (perhaps the browser, perhaps Azure itself, the github issue logs for the problem weren't well documented). So, rather than fetch new ones, `az login` was just trying to reuse the cached version. But of course those creds expire after 6 hours so the process was automatically failing. This change just amounts to forcing Azure to skip one thing in its quest to find working creds. I.e. we know that the default creds are the last thing checked (environment variables first, then azure stuff, then shared cache, etc.), so this skips the thing above it, which happens to be shared cache. It doesn't impact the code at all, but it lets us circumvent any caching that may or may not have happened with login creds. I don't know if anyone else was having this issue (because who else is running a local script with az login for that many hours), but it could've happened as easily as me clicking "yes" when Azure asks if you want it to not ask you for MFA every time. It was so dumb.

src/FunctionApps/python/GenerateCSVs/elr.py
bamader: Wait nevermind I thought you were talking about something else. Yes we absolutely do, added now.

src/FunctionApps/python/GenerateCSVs/fhir.py
DanPaseltiner: Interesting. Okay

src/scripts/analyze_missingness.py
```diff
@@ -67,6 +67,30 @@ def count_records_missing_data(
...
+    how many records our hash actually linked.
+    """
+    prev_hash = None
+    idxs_to_keep = []
+    for row in df.index:
+        if prev_hash == None or df["patientHash"][row] != prev_hash:
+            idxs_to_keep.append(row)
+            prev_hash = df["patientHash"][row]
+    new_df = df.loc[idxs_to_keep, :]
+    new_df.reset_index()
```
bryan-skylight, line 90: Do we need to add a `drop=True` to prevent the old index from getting added as a variable?

src/scripts/analyze_missingness.py
```diff
@@ -92,16 +123,31 @@ def count_records_missing_data(
...
-
-    pre_linkage_covid.drop(
-        columns=[
-            c
-            for c in pre_linkage_covid.columns
-            if c not in list(equity_fields.keys()) + ["patientHash"]
-        ],
-        inplace=True,
+    pre_linkage_elr = pre_linkage_elr.merge(covid_loincs, on="loincCode", how="inner")
+    pre_linkage_ecr = pre_linkage_ecr.merge(covid_loincs, on="loincCode", how="inner")
```
bryan-skylight, line 134: My understanding was that for ECR we were filtering using both LOINC code and vaccine code, but this implies that we're only filtering using the LOINC code. Is that right?

bryan-skylight (approved):
A few nitpicky things, but overall this looks good. Please fix your formatting issues though so that all code checks pass.

src/scripts/analyze_missingness.py
bamader: We can. It doesn't really change the result since we drop all the columns that aren't equity fields anyway, but for clarity and cleanliness yeah we probably should.

src/scripts/analyze_missingness.py
bamader: Yes, that's correct. I misspoke earlier when I said we double merged, my bad. In what I've seen the LOINC is a more granular merging criterion to filter out unrelated data, so we thought this was good for our purposes.

PR #105, opened 2022-05-03
Secure log output

spencer-skylight:
## Summary
Previously, we were sending function app messages to the standard logs, which ended up stored locally on the functionapp's service app (linux container).  This wasn't OK, because the output could contain PII, and the log file location is unencrypted.  

Now, we are returning the HTTP status code and accompanying return value.  The code that calls the conversion function stores the response information in a (secure) blob location.  

## Fixes
Previously, if the number of address elements from an incoming element did not equal the number of elements in the standardized address, it resulted in a address key error.  This has been fixed.

## Testing
Tests have been updated to reflect that successful responses should contain a FHIR bundle (resourceType = "Bundle").  The error response is no longer has special handling for OperationOutcome, so corresponding tests have been removed.

### DIFF
--- a/src/FunctionApps/python/tests/IntakePipeline/conversion_test.py
+++ b/src/FunctionApps/python/tests/IntakePipeline/conversion_test.py
@@ -108,99 +108,9 @@ def test_convert_message_to_fhir_success(mock_fhir_post):
 
 
 @mock.patch("requests.post")
-def test_convert_message_to_fhir_failure(mock_fhir_post):
-    mock_fhir_post.return_value = mock.Mock(
-        status_code=400, json=lambda: {"hello": "world"}
-    )
-
-    message = "MSH|blah|foo|test\nPID|some^text|blah\nOBX|foo||||bar^baz&foobar"
-    response = convert_message_to_fhir(
-        message,
-        "some-filename-0",
-        "Hl7v2",
-        "VXU_V04",
-        "microsofthealth/fhirconverter:default",
-        "some-token",
-        "some-fhir-url",
-    )
-
-    mock_fhir_post.assert_called_with(
-        url="some-fhir-url/$convert-data",
-        json={
-            "resourceType": "Parameters",
-            "parameter": [
-                {"name": "inputData", "valueString": message},
-                {"name": "inputDataType", "valueString": "Hl7v2"},
-                {
-                    "name": "templateCollectionReference",
-                    "valueString": "microsofthealth/fhirconverter:default",
-                },
-                {"name": "rootTemplate", "valueString": "VXU_V04"},
-            ],
-        },
-        headers={"Authorization": "Bearer some-token"},
-    )
-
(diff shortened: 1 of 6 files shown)


src/FunctionApps/python/tests/IntakePipeline/conversion_test.py
```diff
@@ -107,59 +107,10 @@ def test_convert_message_to_fhir_success(mock_fhir_post):
     assert response == {"hello": "world"}
 
 
-@mock.patch("requests.post")
-def test_convert_message_to_fhir_failure(mock_fhir_post):
```
bryan-skylight, line 111: This test shouldn't be deleted. The `convert_message_to_fhir` method can still "fail" (i.e. unsuccessfully convert the message to FHIR), but now returns a non-empty `dict`. This test should be updated to reflect the new return value, not be deleted.

src/FunctionApps/python/tests/IntakePipeline/conversion_test.py
```diff
@@ -107,59 +107,10 @@ def test_convert_message_to_fhir_success(mock_fhir_post):
...
-            "issue": [
-                {
-                    "severity": "fatal",
-                    "code": "code-invalid",
-                    "diagnostic": "error-description",
-                }
-            ],
-        },
-    )
+def test_log_generic_error(mock_log, mock_fhir_post):
```
bryan-skylight, line 112: I think this test is conflating two things, both checking that the logging was properly called while also asserting that the return value from `convert_message_to_fhir` is correct. I'd like to see this split back out so that the tests are more explicit.

src/FunctionApps/python/tests/IntakePipeline/conversion_test.py
```diff
@@ -233,10 +185,13 @@ def test_log_generic_error(mock_log, mock_fhir_post):
     )
 
     mock_log.assert_called_with(
-        "Error during $convert-data -- HTTP Code: 400, Response Content some-error"
+        "HTTP 400 code encountered on $convert-data for some-filename-0"
     )
 
-    assert response == {}
+    assert response == {
```
bryan-skylight, line 191: I don't think we need to assert that the call from `convert_message_to_fhir` is correct here as that's not the point of this test.

spencer-skylight:
> LGTM, although I think given a longer timeline, we'd want to see if we could lock down the logs a bit

@graham-skylight are you talking about making sure the storage that keeps the functionapp logs is encrypted and/or perhaps storing them in a more protected location of the filesystem?  I was thinking about asking DevOps about that...

bryan-skylight (approved):
Thanks for making those changes! If you remove the `assert response == {...}` condition on line 263 of `conversion_test.py` I think this is good to go.

spencer-skylight:
Confirmed the code was good as-is with the function name change. :+1:   Merging.

PR #109, opened 2022-05-13
Dan/modularization refactor

DanPaseltiner:
This PR is a refactor of the IntakePipeline. It moves all logic for the individual steps of the pipeline into `phdi-building-blocks` formerly known as `phdi-transforms`. Efforts have been made to ensure that every transform building block accepts and receives a FHIR bundle. Unit testing coverage and updates have been made in `phdi-building-blocks` and `IntakePipeline`.


celestemespinoza:
Task linked: [CU-29av4a8 Break Up Azure Function to be More Modular](https://app.clickup.com/t/29av4a8)

spencer-skylight (commented):
First off, this looks great.  It seems like we got by with minimal changes - mostly just moving code which is awesome.  

The only thing I might suggest code-wise is splitting blob-related functionality from fhir.py into a new file azure-blob.py.  I could also do this in my forthcoming FHIR export PR as well. Either way I can approve this item as is.

I'd like to get a response to the concern around CI checks when merging to main before adding my +1, but otherwise GTG.

spencer-skylight (commented):
It looks like test_pipeline is missing from `src/FunctionApps/python/tests/IntakePipeline`.  Was that intentional?

src/FunctionApps/python/requirements.txt
```diff
@@ -4,7 +4,7 @@ azure-functions
 azure-identity
 azure-storage-blob
 hl7
-phdi_transforms @ git+https://github.com/CDCgov/prime-public-health-data-infrastructure@main#subdirectory=src/lib/phdi-transforms
+phdi_building_blocks @ git+https://github.com/CDCgov/prime-public-health-data-infrastructure@dan/modularization-refactor#subdirectory=src/lib/phdi-building-blocks
```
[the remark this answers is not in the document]
DanPaseltiner: I think you are right. I'll look at breaking this into 2 PRs.

DanPaseltiner:
> I know the plan is to update this to point to the main branch, but I'm wondering if this will actually work. When we merge, your dan/modularization-refactor branch, it will get automatically deleted, which I assume will cause the post-merge CI checks to fail. I wonder if we need to split this PR into 2 - first to update lib (leaving stuff as-is in IntakePipeline, and a second to update IntakePipeline (and remove phdi_transforms).

I have created a new branch called `dan/TEMP-phdi-building-blocks` that is effectively a copy of this branch. Additionally, I have updated `python/requirements.txt` on this branch to install `phdi_building_blocks` from `dan/TEMP-phdi-building-blocks`. After merging this PR `dan/TEMP-phdi-building-blocks` will persist which I believe addresses the concerns about the merge CI checks. After merging this PR we can follow up with a second one that points `python/requirements.txt` to main. At that point we can delete dan/TEMP-phdi-building-blocks`. This approach seemed a little easier to implement than creating a separate PR containing only changes to `phdi_building_blocks`.

src/lib/phdi-building-blocks/phdi_building_blocks/azure_blob.py
```diff
@@ -0,0 +1,30 @@
...
+    message: str = None,
+) -> None:
+    """
+    Store the given data, which is either a FHIR bundle or an HL7 message in the
+    appropriate output container
+    """
+    client = get_blob_client(container_url)
+    blob = client.get_blob_client(str(pathlib.Path(prefix) / bundle_type / filename))
+    if message_json is not None:
+        blob.upload_blob(json.dumps(message_json).encode("utf-8"), overwrite=True)
```
bamader, line 28: How will this interact with things once the pipeline has already been run once and we just want to process new data? Shouldn't we set `overwrite=False`) and not write the contents if a file of the same name already exists?

src/lib/phdi-building-blocks/phdi_building_blocks/geo.py
```diff
@@ -0,0 +1,119 @@
...
+
+                address["extension"].append(
+                    {
+                        "url": "http://hl7.org/fhir/StructureDefinition/geolocation",
+                        "extension": [
+                            {"url": "latitude", "valueDecimal": geocoded.lat},
+                            {"url": "longitude", "valueDecimal": geocoded.lng},
+                        ],
+                    }
+                )
```
bamader, line 108: This may not be the place for it but I actually just spotted the error that was stopping the metrics tracking from successfully calculating during the one trial run before the demo. If the geocoder fails, then `std_addresses` doesn't get an addition, which means the length of the two arrays could get mismatched. To combat this, we need to do two things:
1. in line 88, instead of setting `std_one_line = ""`, we should have `std_one_line = one_line`.
2. we should remove line 95 and instead call `std_addresses.append(std_one_line)` right before we perform the `any_didffs` calculation.
These changes ensure the lengths of the two arrays are the same and also allow for correct stats tabulation: in the current case, if geocoding fails, we count that as a diff and so we'll get a +1 bump to our number of records that had the address changed. But this is a vacuous stat that we shouldn't get credit for, so instead appending the same raw one line if geocoding fails ensures that it won't count this as the standardization working.

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
```diff
@@ -1,7 +1,7 @@
 import hashlib
 
 
-def add_patient_identifier(salt_str: str, bundle: dict) -> None:
+def add_patient_identifier(bundle: dict, salt_str: str) -> dict:
```
bamader, line 4: Why the switch here?

src/lib/phdi-building-blocks/phdi_building_blocks/name_standardization.py
```diff
@@ -0,0 +1,59 @@
...
+        for name in patient.get("name", []):
+            if "family" in name:
+                std_family = standardize(name["family"])
+                raw_family = name["family"]
+                patient["extension"].append(
+                    {
+                        "url": "http://usds.gov/fhir/phdi/StructureDefinition/family-name-was-standardized",  # noqa
+                        "valueBoolean": raw_family != std_family,
+                    }
+                )
```
bamader, line 42: Something we'll have to consider when we start importing using the FHIR server is that the extensions we're appending for metrics tracking don't actually resolve to valid URLs. Perhaps not a problem, but if a user's data suddenly shows up with some "garbage" / "fake" extensions in there that we used for our purposes that might be seen as losing data quality or a reduction in data cleanliness.

src/lib/phdi-building-blocks/phdi_building_blocks/phone_standardization.py
```diff
@@ -0,0 +1,49 @@
...
+        for telecom in patient.get("telecom", []):
+            if telecom.get("system") == "phone" and "value" in telecom:
+                transformed_phone = standardize(telecom["value"])
+                raw_phones.append(telecom["value"])
+                std_phones.append(transformed_phone)
+                telecom["value"] = transformed_phone
+        any_diffs = len(raw_phones) != len(std_phones) or any(
+            [raw_phones[i] != std_phones[i] for i in range(len(raw_phones))]
+        )
+        patient["extension"].append(
```
bamader, line 43: We should do the same thing here that I recommended for handling address standardization so we don't get artificially higher counts if the system fails to do anything.

bamader (commented):
Seems like a good start. I'm holding off on approval until others review and we've all had sufficient time to discuss each other's thoughts here.

spencer-skylight:
> >It looks like test_pipeline is missing from src/FunctionApps/python/tests/IntakePipeline. Was that intentional?

> This should be resolved now.
Looks good.

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -1,24 +1,28 @@
...
     convert_batch_messages_to_list,
     convert_message_to_fhir,
     get_file_type_mappings,
 )
-from config import get_required_config
-from phdi_transforms.geo import get_smartystreets_client
+
+from phdi_building_blocks.geo import get_smartystreets_client, geocode_patient_address
+from phdi_building_blocks.name_standardization import standardize_patient_name
+from phdi_building_blocks.phone_standardization import standardize_patient_phone
```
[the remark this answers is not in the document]
DanPaseltiner: I had thought about that structure, but thought that depending on how many standardization functions we develop a single module might become awkward. We will also likely want to start standardizing names other than just the patient. If the team would prefer collecting everything into a single module I'm happy to make that change.

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -55,8 +59,11 @@ def run_pipeline(
 
     if response and response.get("resourceType") == "Bundle":
         bundle = response
-        transform_bundle(geocoder, bundle)
-        add_patient_identifier(salt, bundle)
+        standardize_patient_name(bundle)
+        standardize_patient_phone(bundle)
+        geocode_patient_address(bundle, geocoder)
```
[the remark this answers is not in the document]
DanPaseltiner: This is the spot I had in mind this morning. Personally, I like the explicit nature of pass by value, but the function I replaced didn't do that so I left things alone.

src/FunctionApps/python/requirements.txt
```diff
@@ -4,7 +4,7 @@ azure-functions
 azure-identity
 azure-storage-blob
 hl7
-phdi_transforms @ git+https://github.com/CDCgov/prime-public-health-data-infrastructure@main#subdirectory=src/lib/phdi-transforms
+phdi_building_blocks @ git+https://github.com/CDCgov/prime-public-health-data-infrastructure@dan/TEMP-phdi-building-blocks#subdirectory=src/lib/phdi-building-blocks
```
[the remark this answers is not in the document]
DanPaseltiner: Yes indeed. Spencer and I have a conversation going about this.

src/FunctionApps/python/IntakePipeline/__init__.py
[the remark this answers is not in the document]
spencer-skylight: Thinking about this more, I think Dan's goal was to give more granular control over which standardizations should be applied per STLT.  I am leaning towards Brandon's suggestion here of combining for the following reasons:
* One of our goals (I think) is to make data standardization uniform _across_ STLTs, so giving fine grain control over which standardizations to apply would work against that.
* If we wanted to give control over when or how to apply certain standardizations, I'd argue a better way to do that is through configuration of a single building block.  This point is certainly arguable, but that's where I'm leaning right now.

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
DanPaseltiner: I was trying to adopt the convention of always passing the bundle as the first argument for all of bundle transform building blocks. The change to returning a dict is so that we can pass by value.

src/FunctionApps/python/requirements.txt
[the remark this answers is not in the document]
DanPaseltiner: https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/109#issuecomment-1127760743

src/FunctionApps/python/IntakePipeline/__init__.py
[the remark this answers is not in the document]
DanPaseltiner: My thought was that eventually name_standarization might contain functions like standardize_provider_name() and such.

src/lib/phdi-building-blocks/phdi_building_blocks/azure_blob.py
spencer-skylight: This is existing code (just moved to another file).  overwrite=False was the original behavior, and was causing problems when we re-ran files after code updates and wanted those changes to simply overwrite the previous run's data.  For that reason, Graham had suggested we set overwrite=True which seemed like a good idea to me at the time.  It's not the most intelligent solution, and could be improved, but in any case, it's probably out of scope for this PR.

DanPaseltiner: That's probably a good catch. @spencer-skylight I believe you wrote this initially what do you think?

src/lib/phdi-building-blocks/phdi_building_blocks/geo.py
DanPaseltiner: What you have said makes sense, but I would propose making these corrections in a separate PR around tracking for pipeline metrics. While refactoring I was also wondering if we could bundle this tracking into another module that can be called internally from other building blocks. Maybe more trouble than it is worth, but might be nice for developing additional transforms down the road and easily adding logging.

src/lib/phdi-building-blocks/phdi_building_blocks/phone_standardization.py
DanPaseltiner: As above that feels like maybe a separate PR to me.

src/lib/phdi-building-blocks/phdi_building_blocks/name_standardization.py
spencer-skylight: IMO there isn't anything wrong with us defining our own extensions, and I can see information about what data was standardized being valuable to STLTs even beyond metrics collection.  I think it would be good to discuss/reach consensus about if others agree.  We could also discuss if this is the right level to record this info, or whether we wanted, for example, to collect the original value received.  

Here's an example of a vendor (one I worked a lot with in a past life) who publishes a list of extensions.  
https://docs.intersystems.com/components/csp/docbook/fhir_extn_sda3/index.html

Navigating to any of the URLs listed in this page will redirect to the page itself.  If we want to define our own extensions (which I assume we will for one reason or another), we would just need to publish them similarly.

spencer-skylight (approved):

src/FunctionApps/python/IntakePipeline/__init__.py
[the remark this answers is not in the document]
bamader: Spencer I really like those two points you made. In particular, one of the big pieces of feedback we got is that STLTs will be able to more easily adopt our building blocks if they're configurable, so I would definitely push for having a single standardization building block with a larger number of tweakable functions. I think that also helps keep the design principle clean by making standardization a single "chunk" entity that a user has to be cognizant of. So my preference is definitely combining these into a single building block.

src/FunctionApps/python/IntakePipeline/__init__.py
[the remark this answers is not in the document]
bamader: I think whichever way we decide, we should leave that for a separate PR if we do decide to make the change. Right now, things look good as more of a rearrange file systems PR, and changing to pass by value will be enough of a large changeset on its own.

src/lib/phdi-building-blocks/phdi_building_blocks/azure_blob.py
bamader: That's sound reasoning, Spencer. This changeset is big enough as is so probably no need to modify the scope here, especially if things were causing problems the other way.

src/lib/phdi-building-blocks/phdi_building_blocks/geo.py
bamader: Yeah that's totally fair Dan. I figured a reorganization PR isn't the place to do that. It's at least somewhat comforting to know though that we've identified the early candidate for the error-cause in metrics tracking at least. And I think eventually we probably should bundle the tracking separately into another module (perhaps putting it in the `utils.py` building block). That will likely be quite difficult because it'll involve creating a framework rather than tagging for a specific purpose, but either way, it's beyond scope for right now.

src/lib/phdi-building-blocks/phdi_building_blocks/name_standardization.py
bamader: That makes sense, Spencer. I defer to your expertise in this domain because you have a _lot_ more familiarity with how extensions are used in the FHIR server in practice, whereas the only things I know are that the FHIR pages tell me certain things in bold letters 😂 . So if it seems okay to you I'm happy to keep them as is, as well.

src/lib/phdi-building-blocks/phdi_building_blocks/phone_standardization.py
bamader: Yep yep agreed

src/lib/phdi-building-blocks/phdi_building_blocks/name_standardization.py
spencer-skylight: Sounds good!  My hope is that we will all learn a lot about how the FHIR server operates as we move into the future.

src/FunctionApps/python/IntakePipeline/__init__.py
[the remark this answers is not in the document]
DanPaseltiner: Just pushed commit [9290c8f](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/109/commits/9290c8f0ad257239a9169e5679cca216deb773cf )that condenses these modules into `standardize.py`.

bamader (approved):
This is now an approve for me, as most conversations have been handled and issues discussed either here or elsewhere. Still would like to get eyes from Bryan but would also love to merge this in sooner rather than later so I can begin work on the principles-based tech debt refactor ticket Ann's asked me to pick up.

src/FunctionApps/python/IntakePipeline/__init__.py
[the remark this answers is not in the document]
DanPaseltiner: This was addressed in commit [[9290c8f](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/109/commits/9290c8f0ad257239a9169e5679cca216deb773cf)]([9290c8f](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/109/commits/9290c8f0ad257239a9169e5679cca216deb773cf))

src/FunctionApps/python/requirements.txt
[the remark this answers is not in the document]
DanPaseltiner: This PR is on `dan/modularization-refactor` and `dan/TEMP-phdi-building-blocks` is a effectively a copy of this branch. Currently the dependency is pointed to this second branch so that when the PR is merged and this branch is deleted nothing (like integration testing) will brake. Planning to follow up with a PR containing a single line change after merging this PR (#109) to point back to main.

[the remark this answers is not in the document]
DanPaseltiner: Here is where @Spencer and I discussed this. https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/109#issuecomment-1127760743

PR #111, opened 2022-05-16
Fhir server export

spencer-skylight:
# Summary
This PR adds a function to the function app in `src/FunctionApps/python` which supports FHIR server exports as well as underlying library changes.


celestemespinoza:
Task linked: [CU-29av4k7 Create Function that Extracts Data from FHIR Server](https://app.clickup.com/t/29av4k7)

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -125,3 +128,184 @@ def upload_bundle_to_fhir_server(bundle: dict, access_token: str, fhir_url: str)
...
+    access_token: str,
+    fhir_url: str,
+    export_scope: str = "",
+    since: str = "",
+    resource_type: str = "",
+    poll_step: float = 30,
+    poll_timeout: float = 300,
+) -> dict:
+    """Initiate a FHIR $export operation, and poll until it completes.
+    If the export operation is in progress at the end of poll_timeout,
```
bamader, line 143: Yeah what happens if this is the case?

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -125,3 +128,184 @@ def upload_bundle_to_fhir_server(bundle: dict, access_token: str, fhir_url: str)
     except Exception:
         logging.exception("Request to post Bundle failed for json: " + str(bundle))
         return
+
+
+def export_from_fhir_server(
+    access_token: str,
+    fhir_url: str,
```
bamader, line 135: The `fhir.py` file is getting pretty long. Wondering if we should move all of the export functions and functionality into their own file. Thoughts?

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -125,3 +128,184 @@ def upload_bundle_to_fhir_server(bundle: dict, access_token: str, fhir_url: str)
...
+
+    if resource_type:
+        export_url += f"{separator}_type={resource_type}"
+        separator = "&"
+
+    return export_url
+
+
+def __export_from_fhir_server_poll_call(
+    export_url: str, access_token: str
```
bamader, line 220: It's a nit because we pass `poll_url` as the arg in the poll function below, but can we change this parameter name to `poll_url`? I was confused when I was first reading through this because export_url is the string we used to actually kick off the export, whereas poll_url is the status checking URL the server sends back in response.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -125,3 +128,184 @@ def upload_bundle_to_fhir_server(bundle: dict, access_token: str, fhir_url: str)
...
+    :param blob_url: Blob URL location to download from Azure Blob storage
+    :param encoding: encoding to apply to the ndjson content, defaults to "utf-8"
+    :return: Downloaded content wrapped in TextIO
+    """
+    bytes_buffer = io.BytesIO()
+    cred = DefaultAzureCredential()
+    download_blob_from_url(blob_url=blob_url, output=bytes_buffer, credential=cred)
+    text_buffer = io.TextIOWrapper(buffer=bytes_buffer, encoding=encoding, newline="\n")
+    text_buffer.seek(0)
+    return text_buffer
```
bamader, line 311: I'm a little confused about this workflow. So the FHIR server export endpoint sticks all of the patient resources in blob storage (probably silver for us, maybe gold, whatever lol) as ndjson files. Our poller will get a 200 back when that process is complete. At that point, we start downloading the ndjson files and converting them to TextIO. It doesn't seem like we actually need to do this for exporting itself (which seems to terminate the second we get that 200 back), so is the download more about setting up future work on these files (e.g. conversion to parquet)?

bamader (commented):
I think this all makes sense. Just some questions on the workflow and purpose of the individual functions involved.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
spencer-skylight: A TimeoutError will bubble up to the function.  The FHIR server will continue to build the export files, but they will not be processed.  There are a few options for change here...
* We could catch and log the error, return an HTTP error code to the caller, but let the function call succeed.  In this case Azure App Insights would record it as a successful request (I think).
* We could cancel the export job and delete the blob container.  

I opted to keep it simple for this iteration since it was intended to be more of a baby step, but I think taking a closer look at this is a good idea for coming sprints.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
spencer-skylight: Yeah, the download doesn't happen automatically after the export.  The new function does call the 2 processes in sequence (export, download), but they are separable.

spencer-skylight:
I completed my TODO of documenting the high-level process in `src/FunctionApps/python/FhirServerExport/README.md`.  

Waiting to finalize PR 109 before taking this out of draft status.

src/FunctionApps/python/FhirServerExport/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+    return func.HttpResponse(status_code=202)
+
+
+def process_resource(resource: dict):
+    if "resourceType" not in resource or "id" not in resource:
+        raise ValueError("Resource encountered without full information.")
+
+    logging.debug(
+        f"Processing resource {resource['id']} of type {resource['resourceType']}"
+    )
```
DanPaseltiner, line 57: Could we consider separating this functionality out into a separate function (or other implementation)? It seems reasonable that a user might want to be able to export data from the FHIR server and use it in NDJSON format without any additional processing or conversion to parquet.

src/FunctionApps/python/FhirServerExport/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+        export_response = fhir.export_from_fhir_server(
+            access_token=access_token,
+            fhir_url=fhir_url,
+            export_scope=req.params.get("export_scope", ""),
+            since=req.params.get("since", ""),
+            resource_type=req.params.get("type", ""),
+            poll_step=poll_step,
+            poll_timeout=poll_timeout,
+        )
+        logging.debug(f"Export response received: {json.dumps(export_response)}")
```
DanPaseltiner, line 29: Given the potential to request large amounts of data from the FHIR server and the polling we are running under the hood in export_from_fhir_server I'm concerned about Azure Function runtime issues. This is definitely a problem with our current implementation and not the underlying code. If we want to move forward with this as proof of concept I'm okay with that, but we need to be aware of this issue.

DanPaseltiner (changes requested):
This code looks clean and clear. I think the code that is being added to pdhi-building-blocks looks good. As I raised in a comment, I'm concerned about implementation in an Azure Function given the potential for long runtimes. Also, I'd like to see the scope of this work limited to exporting data from the FHIR server and address any down stream transforms (to parquet or otherwise) separately.

src/FunctionApps/python/FhirServerExport/__init__.py
DanPaseltiner: Thanks for making this change.

DanPaseltiner:
My intention was to give this a +1 after separating out the functionality to process the NDJSON files. However, given that this PR is dependent on refactoring being done in PR 109 and there is still at least one semi-significant commit in the works for that PR, it seems prudent to merge PR 109 and then be able to review a cleaner diff in this PR to make sure we aren't missing anything.

src/FunctionApps/python/FhirServerExport/__init__.py
spencer-skylight: Yeah, I think there are definitely concerns around using functions for this since by default they have a relatively short timeout.  For now I'm seeing this as a POC of an implementation strategy, so I think we can discuss this more later.

DanPaseltiner (approved):
LGTM. Thanks for your work on this. Implementing in an Azure Function should be fine for POC and we can change down the road when it becomes necessary.

PR #112, opened 2022-05-16
Add directory looping for bronze-additional-records container

bryan-skylight:
# Description
The `data.tf` file, in addition to specifying the permissions for the storage containers, loops through the subdirectories of each container, which [PR#110](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/110) missed. Josiah has mentioned that this might not be necessary, but for the sake of verbosity and completeness it makes sense to add it in.


bryan-skylight:
This PR is no longer needed as the ADF functionality is working without it. Closing it out.

PR #113, opened 2022-05-16
remove default function app and sync Terraform

JosiahSiegel:
This PR removes the unused "default" function app and updates the Terraform to sync with what currently resides in the "test" environment,


operations/app/terraform/modules/data_factory/datasets.tf
```diff
@@ -5,7 +5,7 @@ resource "azurerm_data_factory_dataset_binary" "pdi_datasa" {
   linked_service_name = azurerm_data_factory_linked_service_azure_blob_storage.pdi_datasa.name
 
   azure_blob_storage_location {
-    container                = "bronze"
+    container                = "bronze/additional-records"
```
bryan-skylight, line 8: It's definitely possible I'm reading this wrong, but I believe `SFTPBinarySink` should still point to `bronze` while `SFTPBinarySink_trial` (which has now been renamed to `Azure_Data_Lake` to be more descriptive) should point to `bronze-additional-records`. Is that right? `bronze/additional-records` is not a container/directory that should exist.

bryan-skylight: Also, starting on line 25, there are resources that are defined that I've since deleted from the Data Factory. `DestinationDataset_mzn` no longer exists. I know it's not part of this PR, but just wanted to point it out in case you wanted to modify that as well.

operations/app/terraform/modules/function_app/apps.tf
```diff
@@ -20,35 +20,43 @@ module "pdi_function_app" {
...
+    SMARTYSTREETS_AUTH_TOKEN               = "@Microsoft.KeyVault(SecretUri=https://${var.resource_prefix}-app-kv.vault.azure.net/secrets/SmartyStreetsAuthToken)"
+    CONTAINER_URL                          = "https://${var.resource_prefix}datasa.blob.core.windows.net/bronze"
+    INTAKE_CONTAINER_URL                   = "https://${var.resource_prefix}datasa.blob.core.windows.net/bronze"
+    INTAKE_CONTAINER_PREFIX                = "decrypted/valid-messages/"
+    OUTPUT_CONTAINER_PATH                  = "processed"
+    CSV_INPUT_PREFIX                       = each.value.CSV_INPUT_PREFIX
+    CSV_OUTPUT_PREFIX                      = each.value.CSV_OUTPUT_PREFIX
+    HASH_SALT                              = "@Microsoft.KeyVault(SecretUri=https://${var.resource_prefix}-app-kv.vault.azure.net/secrets/salt)"
+    FHIR_URL                               = "https://${var.resource_prefix}-fhir.azurehealthcareapis.com"
+    "AzureWebJobs.IntakePipeline.Disabled" = each.value.AzureWebJobs_IntakePipeline_Disabled
```
bryan-skylight, line 53: Does this disable our `IntakePipeline` from running? I assume no, but would appreciate some insight here to help me understand what it's doing.

bryan-skylight (approved):
If my comment on line 8 of `operations/app/terraform/modules/data_factory/datasets.tf` are wrong, then this is good to go. Otherwise, that's the only thing I'd change. Thanks for putting this together!

operations/app/terraform/modules/function_app/apps.tf
JosiahSiegel: Similar to above, I only want to match with the values you are currently using.
The current value is 0, so this will not disable it.
https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/113/files#diff-0ffe207d8985f822098698b3dc16966f2f0182a7cb6d50591a8516cc915ae389R18

PR #117, opened 2022-05-19
Point IntakePipeline to load phdi_building_blocks from main.

DanPaseltiner:
Now that #109 has been merged we can point the path used in `IntakePipeline` to load `phdi_building_blocks` `back to main`.


celestemespinoza:
Task linked: [CU-29av4a8 Break Up Azure Function to be More Modular](https://app.clickup.com/t/29av4a8)

PR #119, opened 2022-05-19
replace devops function app to match format shared by other projects

JosiahSiegel:
This PR replaces the current DevOps function app to match the format used by other projects in order to make installation and updates easier.


DanPaseltiner (approved):
LGTM

PR #121, opened 2022-05-20
Allow specifying container during export

spencer-skylight:
# Summary
Update FHIR export to specify a target container to which export files are written.

Additional changes
* Fix how access token is passed to export function
* Improve logging


spencer-skylight:
Renamed to correct typo in name.

PR #122, opened 2022-05-20
Specify container on export

spencer-skylight:
# Summary
Update FHIR export to specify a target container to which export files are written.

Additional changes
* Fix how access token is passed to export function
* Improve logging
* Update terraform - add new fhir-export container

### DIFF
--- a/src/FunctionApps/python/FhirServerExport/__init__.py
+++ b/src/FunctionApps/python/FhirServerExport/__init__.py
@@ -2,6 +2,7 @@ import azure.functions as func
 import config
 import json
 import logging
+import requests
 
 from phdi_building_blocks import fhir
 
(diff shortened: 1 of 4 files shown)


DanPaseltiner (approved):
Looks good. Can the FHIR server write to an existing container? If not should we think about functionality that checks to see if the specified container already exists and if so prompts the user to chose a new container name?

spencer-skylight:
> Can the FHIR server write to an existing container? If not should we think about functionality that checks to see if the specified container already exists and if so prompts the user to chose a new container name?

Actually the specified container _must_ already exist.  The UUID-named folders will be written within that container for each initiated export request.  If the container doesn't exist yet, the polling endpoint will eventually return a 400 and nothing will happen in terms of file creation.  The fact that a 400 error is returned is logged and the funtionapp request is considered "failed".  If you have any suggestions on a better way to handle this, I'm open to them.  I was actually thinking maybe container should be configured as a functionapp setting rather than accepted as a request parameter.  I should also update the documentation, which I will do after we agree on any potential changes.

DanPaseltiner:
Ahh that works differently than I expected. In this case I agree we should handle provisioning the container at the Function App level, seems more robust than the wait and see approach. Your proposal to make it a config of the function app makes sense to me, if part of each execution is an initial step to check if the container exists and create it if not. If we reasonably think we may want to have the FHIR server write to different containers we could pass the container name as  part of the HTTP request to trigger the function. Personally, I think just adding it to local.settings.json is simplest and if we encounter a situation where we need to more flexibly specify the container we can look into this down the road.

bamader (approved):
This LGTM!

PR #123, opened 2022-05-23
Add updated phone number standardization, supporting functions, and testing

DanPaseltiner:
This PR aims to make our phone number standardization process more robust with a focus on being able to handle international phone numbers that may or may not include country codes.

### DIFF
--- a/src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
+++ b/src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
@@ -1,6 +1,6 @@
 import phonenumbers
 import pycountry
-from typing import Callable, Literal, Generator
+from typing import Callable, Literal, Generator, List
 
 from phdi_building_blocks.utils import find_patient_resources
 
@@ -64,7 +64,7 @@ def standardize_patient_name(
 
 def country_extractor(
     resource: dict, code_type: Literal["alpha_2", "alpha_3", "numeric"] = "alpha_2"
-) -> Generator[str, None, None]:
+) -> List[str]:
     """
     Given a FHIR resource yield a generator containing all of the counries found in the
     resource in a standard form sepcified by code_type.
@@ -74,22 +74,22 @@ def country_extractor(
         code_type (str): A string equal to 'alpha_2', 'alpha_3', or 'numeric' to specify
                          which type of standard country identifier to generate.
     Returns:
-        (generator): A generator to return standardized countries when it is iterated
-                     over.
+        countries (list): A list containing country codes as specified by code_type for
+        each country found in resource.
 
     We currently only handle extracting countries from patient resources, however as
     need arises is the future funcionality can be expanded.
     """
+    countries = []
     resource_type = resource.get("resource").get("resourceType")
     if resource_type == "Patient":
         for address in resource.get("resource").get("address"):
             country = address.get("country")
-            yield standardize_country(country, code_type)
+            countries.append(standardize_country(country, code_type))
+    return countries
 
 
-def standardize_phone(
-    raw: str, resource: dict = None, country_extractor: Callable = country_extractor
-) -> str:
+def standardize_phone(raw: str, countries: List = [None, "US"]) -> str:
     """
     Given a phone number and optionally an associated FHIR resource and country
     extraction function if able to parse the phone number return it in the E.164
@@ -98,13 +98,13 @@ def standardize_phone(
     Phone number parsing process:
     A maximum of three attemtps are made to parse any phone number.
     First, we try to parse the phone number without any additional country information.
-    If this succeeds it must have provided in a standard inernational format which is
-    the ideal case. If the first attempt fails and a FHIR resource associated with the
-    phone number and a country extraction function are provided we make a second attempt
-    to parse the phone number using any countries extracted from the resource. Finally,
-    in the case where the second attempt fails for a resource and country extraction
-    function are not provided we make a final attempt to parse the number assuming it
-    is American.
+    If this succeeds it must have been provided in a standard inernational format which
+    is the ideal case. If the first attempt fails and a FHIR resource associated with
+    the phone number and a country extraction function are provided we make a second
+    attempt to parse the phone number using any countries extracted from the resource.
+    Finally, in the case where the second attempt fails for a resource and country
+    extraction function are not provided we make a final attempt to parse the number
+    assuming it is American.
 
     Args:
         raw (str): Raw phone number to be standardized.
@@ -113,88 +113,68 @@ def standardize_phone(
                          contains information indicating which country the phone number
                          is from.
 
-        country_extractor (Callable): A function to parse a provided resource that
-                                      yields a generator to return the standard 2 letter
-                                      abreviation of each country found in the resource
-                                      when itterated over.
+        countries (list): A list containing 2 letter ISO country codes for each country
+                          extracted from resource of the phone number to be standardized
+                          that might indicate it the phone numbers country of origin.
     Returns:
         standardized (str): The standardized phone number in E.164 format when the raw
                             phone number was succesfully parsed and an emptry string
                             otherwise.
     """
 
-    phone_number = None
+    if countries != [None, "US"]:
+        countries = [None] + countries + ["US"]
 
-    # First Attempt - Assume complete international format provided.
-    try:
-        phone_number = phonenumbers.parse(raw, None)
-    except phonenumbers.phonenumberutil.NumberParseException:
-
-        # Second Attempt - If a resource and country extraction function have been
-        #                  provided attempt to parse using countries found in resource.
-        if resource is not None and country_extractor is not None:
-            for country in country_extractor(resource):
-                try:
-                    phone_number = phonenumbers.parse(raw, country)
-                    break
-                except phonenumbers.phonenumberutil.NumberParseException:
-                    continue
-
-        # Third Attempt - Assume the phone number is American.
-        if phone_number is None:
-            try:
-                phone_number = phonenumbers.parse(raw, "US")
-            except phonenumbers.phonenumberutil.NumberParseException:
-                standardized = ""
-
-    if phone_number is not None:
+    standardized = ""
+    for country in countries:
+        try:
+            standardized = phonenumbers.parse(raw, country)
+        except phonenumbers.phonenumberutil.NumberParseException:
+            continue
+    if standardized != "":
         standardized = str(
             phonenumbers.format_number(
-                phone_number, phonenumbers.PhoneNumberFormat.E164
+                standardized, phonenumbers.PhoneNumberFormat.E164
             )
         )
-
     return standardized
 
 
 def standardize_country(
-    country: str, code_type: Literal["alpha_2", "alpha_3", "numeric"] = "alpha_2"
+    raw_country: str, code_type: Literal["alpha_2", "alpha_3", "numeric"] = "alpha_2"
 ) -> str:
     """
     Given a country return it in a standard form as specified by code_type.
 
     Args:
-        country (str): Country to be standardized.
+        raw_country (str): Country to be standardized.
         code_type (str): A string equal to 'alpha_2', 'alpha_3', or 'numeric' to specify
                          which type of standard country identifier to generate.
 
     Returns:
-        (str): Country in standardized form, or None if unable to standardize.
+        standard_country (str): Country in standardized form, or None if unable to standardize.
     """
-    if len(country) == 3 and country.isalpha():
-        country = pycountry.countries.get(alpha_3=country)
-    elif len(country) == 2 and country.isalpha():
-        country = pycountry.countries.get(alpha_2=country)
-    elif len(country) == 3 and country.isnumeric():
-        country = pycountry.countries.get(numeric=country)
-    elif len(country) > 4 and country.isalpha:
-        possible_countries = pycountry.countries.search_fuzzy(country)
+    standard_country = ""
+    if len(raw_country) == 3 and raw_country.isalpha():
+        standard_country = pycountry.countries.get(alpha_3=raw_country)
+    elif len(raw_country) == 2 and raw_country.isalpha():
+        standard_country = pycountry.countries.get(alpha_2=raw_country)
+    elif len(raw_country) == 3 and raw_country.isnumeric():
+        standard_country = pycountry.countries.get(numeric=raw_country)
+    elif len(raw_country) >= 4 and all(x.isalpha() or x.isspace() for x in raw_country):
+        possible_countries = pycountry.countries.search_fuzzy(raw_country)
         if len(possible_countries) == 1:
-            country = possible_countries[0]
-        else:
-            country = ""
-    else:
-        country = ""
+            standard_country = possible_countries[0]
 
-    if country != "":
+    if standard_country != "":
         if code_type == "alpha_2":
-            country = country.alpha_2
+            standard_country = standard_country.alpha_2
         elif code_type == "alpha_3":
-            country = country.alpha_3
+            standard_country = standard_country.alpha_3
         elif code_type == "numeric":
-            country = country.numeric
+            standard_country = standard_country.numeric
 
-    return country
+    return standard_country
 
 
 def standardize_patient_phone(
@@ -212,9 +192,10 @@ def standardize_patient_phone(
         # Transform phone numbers
         raw_phones = []
         std_phones = []
+        countries = country_extractor(resource)
         for telecom in patient.get("telecom", []):
             if telecom.get("system") == "phone" and "value" in telecom:
-                transformed_phone = standardize(telecom["value"], resource)
+                transformed_phone = standardize(telecom["value"], countries)
                 raw_phones.append(telecom["value"])
                 std_phones.append(transformed_phone)
                 telecom["value"] = transformed_phone
--- a/src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
+++ b/src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
@@ -1,6 +1,6 @@
 import phonenumbers
 import pycountry
-from typing import Callable, Literal, Generator, List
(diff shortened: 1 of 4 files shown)


src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -59,18 +62,139 @@ def standardize_patient_name(
...
+    if resource_type == "Patient":
+        for address in resource.get("resource").get("address"):
+            country = address.get("country")
+            yield standardize_country(country, code_type)
+
 
-    >>> standardize_phone("(555) 555-1212")
-    '5555551212'
+def standardize_phone(
+    raw: str, resource: dict = None, country_extractor: Callable = country_extractor
```
bryan-skylight, line 91: Does `country_extractor` need to be passed in like this? Or can we just call it from within `standardize_phone`?

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -59,18 +62,139 @@ def standardize_patient_name(
...
+    Returns:
+        (generator): A generator to return standardized countries when it is iterated
+                     over.
+
+    We currently only handle extracting countries from patient resources, however as
+    need arises is the future funcionality can be expanded.
+    """
+    resource_type = resource.get("resource").get("resourceType")
+    if resource_type == "Patient":
+        for address in resource.get("resource").get("address"):
```
bryan-skylight, line 85: It seems unlikely to me that any given patient resource will have so many addresses that maintaining a list of the countries will be an expensive task requiring a generator. I like this approach, but by just getting a list of countries, you can call this function outside of `standardize_phone` and pass in the list of countries instead of the entire resource.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -59,18 +62,139 @@ def standardize_patient_name(
...
     """
 
-    raw = [x for x in raw if x.isnumeric()]
-    raw = "".join(raw)
-    if len(raw) != 10:
-        raw = None
-    return raw
+    phone_number = None
+
+    # First Attempt - Assume complete international format provided.
```
bryan-skylight, line 128: If you pass in the list of countries, you could then add a line like `countries = [None] + countries + ["US"]`, and then do the following:

```python
standardized = ""
for country in countries:
    try:
        phone_number = phonenumbers.parse(raw, country)
    except phonenumbers.phonenumberutil.NumberParseException:
        continue
if standardized != "":
    standardized = str(
        phoenumbers.format_number(
            phone_number, phonenumbers.PhoneNumberFormat.E164
        )
    )
return standardized
```

This would handle all three of your edge cases in a much more succinct manner.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -59,18 +62,139 @@ def standardize_patient_name(
...
+        country = pycountry.countries.get(alpha_3=country)
+    elif len(country) == 2 and country.isalpha():
+        country = pycountry.countries.get(alpha_2=country)
+    elif len(country) == 3 and country.isnumeric():
+        country = pycountry.countries.get(numeric=country)
+    elif len(country) > 4 and country.isalpha:
+        possible_countries = pycountry.countries.search_fuzzy(country)
+        if len(possible_countries) == 1:
+            country = possible_countries[0]
+        else:
```
bryan-skylight, line 184: If you move the instantiation of `country` to line 174, you can avoid these `else` statements.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -59,18 +62,139 @@ def standardize_patient_name(
...
+        (str): Country in standardized form, or None if unable to standardize.
+    """
+    if len(country) == 3 and country.isalpha():
+        country = pycountry.countries.get(alpha_3=country)
+    elif len(country) == 2 and country.isalpha():
+        country = pycountry.countries.get(alpha_2=country)
+    elif len(country) == 3 and country.isnumeric():
+        country = pycountry.countries.get(numeric=country)
+    elif len(country) > 4 and country.isalpha:
+        possible_countries = pycountry.countries.search_fuzzy(country)
```
bryan-skylight, line 181: I don't think we should be adding in fuzzy matching for countries. Especially not with a way to set a threshold, and certainly not without indicating in the doc string that it's an option.

bryan-skylight (changes requested):
I really love the robustness of this code, but I think some refactors are necessary before it can be accepted.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
spencer-skylight: Looking at the code, I don't think there is a threshold option for fuzzy matching.  Maybe instead you could accept a list or tuple of input formats to check for?
```
input_formats: Iterable[str]
```
input_formats could be defaulted to ("alpha_2", "alpha_3", "numeric", "fuzzy") in the function, and your cases would check for presence of the appropriate value before trying that kind of conversion.

Also, change `code_type` to `output_format`

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -59,18 +62,139 @@ def standardize_patient_name(
...
+    Returns:
+        (str): Country in standardized form, or None if unable to standardize.
+    """
+    if len(country) == 3 and country.isalpha():
+        country = pycountry.countries.get(alpha_3=country)
+    elif len(country) == 2 and country.isalpha():
+        country = pycountry.countries.get(alpha_2=country)
+    elif len(country) == 3 and country.isnumeric():
+        country = pycountry.countries.get(numeric=country)
+    elif len(country) > 4 and country.isalpha:
```
spencer-skylight, line 180: Should this be `country.isalpha()`?  Also what about 4-character country names like Chad or Iran?

spencer-skylight (changes requested):

spencer-skylight:
Great work - this seems to be very robust!  I had a few suggestions to clean it up a bit, some of which wasn't introduced by your work.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -59,18 +62,139 @@ def standardize_patient_name(
...
+    Given a FHIR resource yield a generator containing all of the counries found in the
+    resource in a standard form sepcified by code_type.
+
+    Args:
+        resource (dict): A FHIR resource respresented as a dictionary.
+        code_type (str): A string equal to 'alpha_2', 'alpha_3', or 'numeric' to specify
+                         which type of standard country identifier to generate.
+    Returns:
+        (generator): A generator to return standardized countries when it is iterated
+                     over.
```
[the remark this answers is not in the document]
bamader: Hard agree. We have a fairly well established format of doc-strings at this point, and that's what I'm incorporating into the refactor based on principles as well. The first answer in this stack overflow post https://stackoverflow.com/questions/9195455/how-to-document-a-method-with-parameters has examples of all the kinds I've used so far.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
bamader: Seconded. I think at this point we really need to be emphasizing human-readability, uniformity in design, and composability (can you tell I'm deep in the refactors lol). I don't think we need to provide a generator to the list of countries and think a list that could be accessed outside `standardize_phone` would be much better.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
bamader: I would actually pretty strongly argue we _shouldn't_ be passing functions as `Callable`'s like this. Our principles emphasize composability, and the conversations we've had around how a user should be able to use or not use building blocks to their liking has been of the form 

```
data_1 = building_block_a(raw_data)
data_2 = building_block_c(data_1) # I skipped building_block_b
data_3 = building_block_m(data_2) # skipping BB's c through l
```

That pattern is currently being strongly established in our codebase and in my mind at least it creates some strong benefits around human readability and standardizing what a user can expect. To me it's better to not invoke a callable like this, particular for the smaller case of extracting a country code. In the future, if we do want to add functionality for different kinds of extractors or standardizers, I think those should be given as different functions entirely to keep things discrete, modular, and accessible in bite-sized chunks for non-developer users.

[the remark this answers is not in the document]
bamader: I'm actually working on cleaning this up now Spencer

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
bamader: Very much agree to no fuzzy matching for this. I think as a general rule we should save anything that might use fuzzy matching for post 6/30.

bamader (changes requested):

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
DanPaseltiner: I agree, I was looking for away to only actually loop over the addresses in a patient resource once and opposed to twice. The loops won't be expensive and I agree it would clean up other elements of the code.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
DanPaseltiner: I must have been sleeping for this last condition. There are no country names shorter than 4 characters to it should be `len(country) >= 4 and country.isalpha():`. Thanks for catching this.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
DanPaseltiner: This is great. I'm going to slightly expand on this by setting the default value of countries to `[None, "US"]`. THis maintains the option for the user to not extract and provide a list of countries. In the case where the user does provide a list we'll do `countries = [None] + countries + ["US"]` as you suggest.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
DanPaseltiner: done.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
bryan-skylight: Also, note that I forgot to include a `break` statement before the `except`. That'll be necessary since you want to dump out of the loop upon a successful parsing.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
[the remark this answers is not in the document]
DanPaseltiner: [59b862b](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/123/commits/59b862b89d61b1008946240041079380c18f1fc1) replaces the fuzzy matching with explicit searches on `name `and `official name`. This will let us make more than a good faith effort to handle cases where the FHIR bundle doesn't have an ISO country code, but avoids fuzzy matching.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -59,18 +62,121 @@ def standardize_patient_name(
...
-    if len(raw) != 10:
-        raw = None
-    return raw
+    if countries != [None, "US"]:
+        countries = [None] + countries + ["US"]
+
+    standardized = ""
+    for country in countries:
+        try:
+            standardized = phonenumbers.parse(raw, country)
```
bryan-skylight, line 121: Need to add a `break` statement here. I forgot it in my example code.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -59,18 +62,121 @@ def standardize_patient_name(
...
+                            phone number was succesfully parsed and an emptry string
+                            otherwise.
     """
 
-    raw = [x for x in raw if x.isnumeric()]
-    raw = "".join(raw)
-    if len(raw) != 10:
-        raw = None
-    return raw
+    if countries != [None, "US"]:
```
bryan-skylight, line 115: Q: do you think this would be more readable for the user if we did something like the following?

```python
def standardize_phone(raw: str, countries: List = []) -> str:
    <code>
    countries = [None] + countries + ["US"]
    <code>
```

The thought here is that it reduces the need for the `if` statement. I'm torn on whether or not it makes the code more readable though. I think either way is fine, just wanted to get your read on it.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -59,18 +62,121 @@ def standardize_patient_name(
...
+        standardize.
+    """
+    standard_country = ""
+    if len(raw_country) == 3 and raw_country.isalpha():
+        standard_country = pycountry.countries.get(alpha_3=raw_country)
+    elif len(raw_country) == 2 and raw_country.isalpha():
+        standard_country = pycountry.countries.get(alpha_2=raw_country)
+    elif len(raw_country) == 3 and raw_country.isnumeric():
+        standard_country = pycountry.countries.get(numeric=raw_country)
+    elif len(raw_country) >= 4 and all(x.isalpha() or x.isspace() for x in raw_country):
```
bryan-skylight, line 152: Is the `all(...)` condition necessary? What happens with `pycountry.countries.get(name=raw_country)` if it contains non-alpha characters? I'd assume it just wouldn't match anything and so nothing would be returned? I suppose I'd ask that for all of the extra conditions. I just did some testing with `pycountry`, and it seems to handle these scenarios just fine. I think the code would be much cleaner if we did the following, starting on line 158:

```python
raw_country = raw_country.strip().upper()
standard_country = None
if len(raw_country) == 2:
    standard_country = pycountry.countries.get(alpha_2=raw_country)
if len(raw_country) == 3: # no need for elif since the conditions are mutually exclusive
    standard_country = pycountry.countries.get(alpha_3 = raw_country) or pycountry.countries.get(numeric=raw_country)
if len(raw_country) >= 4:
    standard_country = pycountry.countries.get(name=raw_country) or pycountry.countries.get(official_name=raw_country)
    
if standard_country is not None:
    <code>
```

bryan-skylight: nit: I'd also consider dropping the `_country` suffix from the variable name as it should be clear from the function what `raw` and `standard` would represent, and I'd also recommend changing `standard` to `standardized`.

bryan-skylight (changes requested):
The need for a `break` statement is the only blocking comment in this review.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
DanPaseltiner: That's a carry over from addressing one of Spencer's comments before eliminating the fuzzy matching. Yeah, it isn't needed any more. I'm not inclined to eliminate the use of `elif` though. I agree the conditions are mutually exclusive, but Python doesn't know that and will evaluate all the conditions as opposed to breaking out of the block early after one condition is met.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
DanPaseltiner: Addressed in [d06eb70](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/123/commits/d06eb7053c986c3b132e9e7dab26762050852af1).

spencer-skylight (approved):
Looks good to me, with recent changes!

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -59,18 +62,109 @@ def standardize_patient_name(
...
+    :return str standard: Country in standardized form, or None if unable to
+    standardize.
+    """
+    standard = ""
+    raw = raw.strip().upper()
+    if len(raw) == 3:
+        standard = pycountry.countries.get(alpha_3=raw)
+    elif len(raw) == 2:
+        standard = pycountry.countries.get(alpha_2=raw)
+    elif len(raw) == 3:
```
bryan-skylight, line 152: This `if` logic looks off as you're testing `len(raw) == 3` twice. Is that intentional?

bryan-skylight (changes requested):
Blocking merging as I think the `if` logic needs to be updated before merging.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
DanPaseltiner: Nope, not intentional just a result of working too fast. I think the changes in [d5edfd9](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/123/commits/d5edfd93601691afb236388136e517254d8a4980) are a little more readable and resolve this issue.

spencer-skylight:
> see in ISO-3166-1 numbers are always padded with zeros to be 3 digits, but I'm wondering if we should handle 1 and 2 digit numeric codes?
Discussed with @DanPaseltiner and we decided it is safer not to try making this correction on behalf of the sending party.  We would be making too many assumptions around what is happening with the data.

celestemespinoza:
Task linked: [CU-2j2mpqu Use phonenumbers lib to standardize numbers](https://app.clickup.com/t/2j2mpqu)

bryan-skylight (approved):
LGTM! Nice work!

bamader (approved):
Good stuff! I appreciate the full docstrings, as that means when I resolve merge conflicts with this and my PR I just have to change some of the naming and it should be pretty seamless.

PR #124, opened 2022-05-23
Configure container in functionapp

spencer-skylight:
* Rather than allowing the target container for export files to be specified as part of the export request, configure it in the function app serttings.
* Add new fhir-exports default export storage container to terraform

### DIFF
--- a/src/FunctionApps/python/FhirServerExport/__init__.py
+++ b/src/FunctionApps/python/FhirServerExport/__init__.py
@@ -13,6 +13,10 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
     poll_step = float(config.get_required_config("FHIR_EXPORT_POLL_INTERVAL", 30))
     poll_timeout = float(config.get_required_config("FHIR_EXPORT_POLL_TIMEOUT", 300))
 
+    container = config.get_required_config("FHIR_EXPORT_CONTAINER", "fhir-exports")
+    if container == "<none>":
+        container = ""
+
     cred_manager = fhir.AzureFhirserverCredentialManager(fhir_url=fhir_url)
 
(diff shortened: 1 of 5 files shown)


src/FunctionApps/python/FhirServerExport/__init__.py
```diff
@@ -13,6 +13,10 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
     poll_step = float(config.get_required_config("FHIR_EXPORT_POLL_INTERVAL", 30))
     poll_timeout = float(config.get_required_config("FHIR_EXPORT_POLL_TIMEOUT", 300))
 
+    container = config.get_required_config("FHIR_EXPORT_CONTAINER", "fhir-exports")
+    if container == "<none>":
```
bryan-skylight, line 17: Is there a reason for setting the empty value to `"<none>"` as opposed to just `""`, which is what you set it to in the code?

spencer-skylight: Yes, that would have been my preference and I did test the following approach:
* Un-set maps to the default fhir-exports container name
* Specifying an empty string maps to no container

Unfortunately, our `get_required_config` implementation doesn't differentiate between an un-set variable, and empty string (at least when run locally).  Both end up returning the default value (fhir-exports).  Rather than figuring out a change to `get_required_config` to this PR I decided to work around it by introducing the `<none>` special value.

bryan-skylight (approved):
LGTM

celestemespinoza:
Task linked: [CU-2y7qhfu Implement additional export endpoints in FHIR server extraction](https://app.clickup.com/t/2y7qhfu)

bamader (approved):
LGTM! Nice, short, and easy.

PR #125, opened 2022-05-24
Enforce Human-Centered Design Principles

bamader:
This changeset enforces a strict set of human-centered design principles in our codebase. We emphasize human clarity / readability first and foremost, but we also value consistency, composability, and accessibility. A large percentage of changes in this PR concern doc strings, which have now been assigned a standardized format in our codebase. However, there are some modularity changes to function logic as well, including breaking out repeated code into its own function or abstracting confusing pieces of logic or code away to enhance human readability.

In particular, I'd like people's thoughts on my decision choice around "modes" of a function (standardizing is a good example, for instance). Previously, we had one large function that just handled bulk standardization of all patients in a bundle. I've now broken this out into discrete helpers that each build on each other: one function to standardize one instance of, e.g. a name; one function to standardize all names for one patient; and one function to standardize all patients in a given bundle. Providing three levels of granularity there seemed like a good way to increase the accessibility of our building blocks by exposing more ways to employ them in STLTs own functionality. Further, the way I've implemented what _types_ of standardization can be done inside of a bundle is now different. When we had only the one large function, a user had to pass in a Callable that would represent their standardization function of choice. Now, the highest-level function (all patients in a bundle) accepts a string instead denoting which of several possible standardization options a user might wish to use (currently we only have one implemented, but this could grow later on). There's then simple if/else logic to see which callable the _helper methods_ should invoke. To me, this actually achieves the best balance between the principles I'm championing here and flexibility in code. Overall, we haven't changed the ability for a user to alter the kinds of standardization or transforms they're performing; now, instead of importing another function and passing in a callable (which I would argue is not HCD, as non-dev or more general users might not understand that relationship, and it's also an extra import hassle), the user passes in a string with the name of the operation they want to perform (I would argue that most users should immediately be able to intuitively understand naming different options). See if this makes sense to you all as well. If people feel strongly about using just callables, I'll switch it back, but trying to really internalize these principles over the last few days I think some string simplification on the front-end without complicating the back-end is a worthwhile trade.


bamader:
The IntakePipeline tests need some work now that we're returning by value, so I'll pick that back up tomorrow

src/lib/phdi-building-blocks/tests/test_geo.py
```diff
@@ -108,4 +108,4 @@ def test_geocode_patient_address(patched_geocoder):
     )
     patched_geocoder.return_value = geocoded_response
 
-    assert geocode_patient_address(raw_bundle, mock.Mock()) == standardized_bundle
+    assert geocode_patients_in_bundle(raw_bundle, mock.Mock()) == standardized_bundle
```
DanPaseltiner, line 111: Looks like we are missing a unit tests for `standardize_addresses_for_patient` and ` geocode_single_address`. I realize they are tested implicitly through the test for `geocode_patients_in_bundle`, however if we are splitting this functionality up into separate functions an end user might choose to use them outside of our functions that wrap them so I think we should write some explicit tests.

bamader:
Code is now updated to reflect integration with Dan's great work on country codes. I wound up making just a couple small changes to the extract countries signature to specifically accept only a patient resource rather than a general "resource". In the future we very well might want to support more types, but until such a time I think this current implementation best fits in the HCD architecture I've designed and best integrates with the standardization modes type work I've done. There's no changed functionality, I just want us to emphasize completeness on what's there before we get too ahead of ourselves making things general for the future. When we are ready to include countries from non-patient resources, it should be a simple matter of defining some more helper functions and just plugging them into the base signature.

src/lib/phdi-building-blocks/tests/test_linkage.py
```diff
@@ -109,7 +112,7 @@ def test_add_patient_identifier():
         "use": "temp",
     }
```
DanPaseltiner, line 113: I think our unit tests will be more readable if we move this "testing data" into JSON files.

src/lib/phdi-building-blocks/tests/test_linkage.py
```diff
@@ -1,4 +1,7 @@
-from phdi_building_blocks.linkage import generate_hash_str, add_patient_identifier
```
DanPaseltiner, line 1: Are we missing a test for `get_field_with_use`?

src/lib/phdi-building-blocks/tests/test_standardize.py
```diff
@@ -4,18 +4,19 @@
```
DanPaseltiner, line 5: Similar to my comment on `test_geo.py` I think we should add explicit tests for all functions, even ones we wrap inside others that are tested.

src/lib/phdi-building-blocks/phdi_building_blocks/geo.py
```diff
@@ -1,119 +1,193 @@
...
-                            {"url": "longitude", "valueDecimal": geocoded.lng},
-                        ],
-                    }
-                )
-        any_dffs = (len(raw_addresses) != len(std_addresses)) or any(
-            [raw_addresses[i] != std_addresses[i] for i in range(len(raw_addresses))]
+    return (geocoded_result, address_is_different)
+
+
+def standardize_addresses_for_patient(
```
bryan-skylight, line 134: I'm having a hard time following all of the various functions that are getting called for our geocoding. We're calling functions `standardize.*` when they're also geocoding, and geocoders are returning more than the geocoded results. I think this all needs to be reworked to make it more readable and intuitive.

src/lib/phdi-building-blocks/phdi_building_blocks/geo.py
```diff
@@ -1,119 +1,193 @@
...
+    :param dict patient: A Patient resource FHIR profile
+    :param us_street.Client client: The API client to geocode with
+    :param bool add_metrics_extension: Whether to add tracking metrics into the
+    patient resource denoting that one or more addresses was standardized.
+    :return: None
+    """
+
+    # We can track whether geocoding improved addresses by taking the logical
+    # OR everytime we geocode a new address
+    any_address_was_changed = False
```
bryan-skylight, line 151: I know we talked about adding in these "trackers" for the purposes of demonstrating if our functions were having an impact on the data, but I don't think it should be part of our production code. In reality, we should have a notebook that reads a sample of files from the raw storage container, runs the data through our transformations using our SDK, and checks to see if the values changed. This should _not_ be part of production code. 

Happy to hear dissenting opinions.

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
```diff
@@ -1,68 +1,100 @@
 import hashlib
+import copy
+from phdi_building_blocks.utils import find_patient_resources, get_one_line_address
 
 
-def add_patient_identifier(bundle: dict, salt_str: str) -> dict:
+def add_linking_identifier_to_patients_in_bundle(
```
bryan-skylight, line 6: I think `add_patient_identifer` is the more descriptive function name.

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
```diff
@@ -1,68 +1,100 @@
...
+    bundle: dict, salt_str: str, overwrite: bool = True
+) -> dict:
     """
-    Given a FHIR resource bundle (defined as a dictionary of resources
-    containing at least one patient resource):
-      - identify the patient resource(s) in the bundle
+    Given a FHIR resource bundle:
+      - identify all patient resource(s) in the bundle
       - extract standardized name, DOB, and address information for each
```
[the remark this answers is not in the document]
bryan-skylight: I don't think we're "enforcing" the use of other building blocks, but the docstring does imply they've been used, which as you point out isn't necessarily true.

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
```diff
@@ -1,68 +1,100 @@
...
+        patient = resource.get("resource")
 
-            # Combine given and family name
-            recent_name = next(
-                (name for name in patient["name"] if name.get("use") == "official"),
-                patient["name"][0],
-            )
+        # Combine given and family name
+        recent_name = get_field_with_use(patient, "name", "official", 0)
+        name_parts = recent_name.get("given", []) + [recent_name.get("family")]
```
bryan-skylight, line 35: What if the `family` name isn't provided? Would this result in `name_parts = [] + [None]`? Is the `given` name _always_ in a list?

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
```diff
@@ -1,68 +1,100 @@
...
-                    "value": hashcode,
-                    # Note: this system value corresponds to the FHIR specification
-                    # for a globally used / generated ID or UUID--the standard here
-                    # is to make the use "temporary" even if it's not
-                    "system": "urn:ietf:rfc:3986",
-                    "use": "temp",
-                }
-            )
+
+def get_field_with_use(patient: dict, field: str, use: str, default_field: int) -> str:
```
bryan-skylight, line 64: This function name is not intuitive to me, but that could be because I'm not fully familiar with HL7. Will "use" be an easily recognized term by people working with public health data and FHIR? 

I'd also change the `patient` parameter to something like `resource` since this code can likely be used for other `resources`. `data` might even be a better name for it, with a description of what it represents in the docstring.

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
```diff
@@ -1,68 +1,100 @@
...
+    :param dict patient: Patient from a FHIR bundle
+    :param str field: The field to extract
+    :param str use: The use the field must have to qualify
+    :param int default_field: The index of the field type to treat as
+    the default return type if no field with the requested use case is
+    found
+    :return: The requested use-case-type field
+    :rtype: str
+    """
+    return next(
```
bryan-skylight, line 81: This needs a comment to explain what it's doing.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -1,93 +1,172 @@
...
-    'JOHN DOE'
+def non_numeric_caps_standardization(raw_name: str) -> str:
     """
-
-    raw = [x for x in raw if not x.isnumeric()]
-    raw = "".join(raw)
-    raw = raw.upper()
-    raw = raw.strip()
-    return raw
+    Perform standardization on a provided name string. Name
```
bryan-skylight, line 10: We've changed the name of the function, but the docstring indicates it's still just for names. If we're trying to generalize this to handle any type of text, then I recommend we do something like the following:

```python
def standardize_text(
    text: str, 
    remove_numbers=False, 
    case=Literal["upper", "lower", "title", "original"]="original",
    <other parameters>
) -> str:
    <code>
```
And then we remove all references to name and anything else, and we move it to `utils.py`. Hell, there are probably already functions like this that we could just import.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -1,93 +1,172 @@
...
+    raw_name = raw_name.upper()
+    raw_name = raw_name.strip()
+    return raw_name
 
 
-def standardize_patient_name(
-    bundle: dict, standardize: Callable = standardize_name
+def standardize_patient_names_in_bundle(
+    bundle: dict,
+    standardization_mode: str = "non_numeric_caps",
```
bryan-skylight, line 29: I'm not a big fan of this approach for two reasons:

1. It requires the end user to familiarize themselves with our terminology, increasing the cognitive load on their end, and
2. it limits the end user to only the combination of transformations that have been pre-coded.

Instead, if we provided parameters like the ones I listed above under the `non_numeric_caps_standardization` function, then the user gets all of the control they need with minimal cognitive overhead.

What are your thoughts on that?

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -1,93 +1,172 @@
...
+    add_name_metrics: bool = True,
```
bryan-skylight, line 30: Again, I'm _really_ opposed to including this in our production code.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -1,93 +1,172 @@
...
+    :param str standardization_mode: The type of standardization to perform.
+    Options include: "non_numeric_caps"
+    :param bool add_name_metrics: Whether to store tracking data about
+    if any names were changed in the patient resources
+    :param bool overwrite: Whether to overwrite the original data with
+    the transformed values. Default is yes.
+    :return: The bundle with all patients having transformed names
+    :rtype: dict
+    """
+    # Copy the data if we don't want to overwrite the origina
```
bryan-skylight, line 51: Typo: `origina` vs `original`

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -1,93 +1,172 @@
...
-                        "url": "http://usds.gov/fhir/phdi/StructureDefinition/given-name-was-standardized",  # noqa
-                        "valueBoolean": any_diffs,
-                    }
-                )
-                name["given"] = std_givens
 
-    return bundle
+def standardize_names_for_patient(
+    patient: dict,
+    transform_func: Callable = non_numeric_caps_standardization,
```
bryan-skylight, line 69: We should not be passing functions into functions. All of these functions are globally accessible, and passing functions in just makes the code harder to read in my opinion. Instead, we could have a `FUNC_MAP` global variable that looked something like

```python
FUNC_MAP = {
    "non_numeric_uppercase": non_numeric_uppercase_standardization,
}
```
and then in the code we have:

```python
def standardize_names_for_patient(patient: dict, transformation: str, ...):
    transform_func = FUNC_MAP[transformation] # note that we're okay with a KeyError exception here
    <code>
```

bryan-skylight: But again, I don't think we should be using this approach, and instead should add the various aspects of the data they can change as parameters to the function.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -1,93 +1,172 @@
...
+            transformed_name = transform_func(name.get("family", ""))
+            family_name_altered = family_name_altered or (
+                transformed_name != name.get("family", "")
+            )
+            name["family"] = transformed_name
+
+        # Given names are stored in a list, as there could be multiple,
+        # so process them all and take the overall diff for metrics
+        if "given" in name:
+            transformed_names = [transform_func(g) for g in name.get("given", [])]
```
bryan-skylight, line 103: If we used the parameters as described above, this line would change to

```python
transformed_names = [standardize_text(g, <parameters>) for g in name.get("given", [])]
```

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -1,93 +1,172 @@
...
+            "url": "http://usds.gov/fhir/phdi/StructureDefinition/family-name-was-standardized",  # noqa
+            "valueBoolean": family_name_was_changed,
+        }
+    )
 
-    We currently only handle extracting countries from patient resources, however as
-    need arises is the future funcionality can be expanded.
 
-    :param dict resource: A FHIR resource respresented as a dictionary.
+def country_extractor(
```
bryan-skylight, line 149: I think we should leave this function generalized for all resources (I can see a scenario where a user would want to standardize the phone number for a Physician), pass in the full bundle, allow the end user to specify the resource they want to target, and use the `find_<x>_resources()` function that's been defined to narrow down to just that. 

We're effectively taking a step backward with this change by making it less generalized than it already was.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -167,35 +246,104 @@ def standardize_country(
...
 
 
-def standardize_patient_phone(
-    bundle: dict, standardize: Callable = standardize_phone
-) -> dict:
-    """Given a FHIR bundle and a standardization function, standardize the phone numbers
-    in all patient resources in the bundle. By default the standardize_phone function
-    defined in this module is used, but the user is free to provide their own function
-    specifying a standardization process as long as it accepts and returns a string."""
+def phone_truncation_standardization(raw_phone: str, *args) -> str:
```
bryan-skylight, line 249: Do we need this function now that we have phone standardization?

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -167,35 +246,104 @@ def standardize_country(
...
     for resource in find_patient_resources(bundle):
         patient = resource.get("resource")
+        if standardization_mode == "truncation":
+            standardize_phone_numbers_for_patient(
+                patient, phone_truncation_standardization, add_phone_metrics
+            )
+        elif standardization_mode == "country":
+            standardize_phone_numbers_for_patient(
+                patient, phone_country_standardization, add_phone_metrics
+            )
```
bryan-skylight, line 302: I think this is another good reason for not going with this approach. Any time we add a method of standardization, we have to update it in a _bunch_ of places, which makes the code harder to read. And here, we could end up with a long list of `if/elif` statements. If we end up using this approach of specifying a helper function to perform the standardization, then we absolutely need to make use of the `FUNC_MAP` variable.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -167,35 +246,104 @@ def standardize_country(
...
+        else:
+            raise ValueError("Invalid standardization mode supplied.")
```
bryan-skylight, line 304: I'm wondering if maybe we want to avoid raising exceptions in scenarios like this. If an invalid standardization is specified, maybe we just return the raw data and print that no standardizations were performed due to misspecification?

src/lib/phdi-building-blocks/phdi_building_blocks/utils.py
```diff
@@ -2,9 +2,30 @@
 
 
 def find_patient_resources(bundle: dict) -> List[dict]:
```
bryan-skylight, line 4: I think this function should be generalized such that it's `find_resource_by_type(bundle: dict, resource:str) -> List[dict]`.

bryan-skylight (changes requested):
I think this is a really great start, and I'm really appreciative of you taking it on! The work you've done has definitely highlighted some areas where I think we as a team need to have a discussion and decide how we want to handle things.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -4,83 +4,54 @@
...
+    """
+    Strip the file type suffix from the blob name, and instead
+    append the message index.
+    :param str blob_name: The name of the blob to modify
+    :param int message_index: The index of this message in the batch
+    :return: The derived filename
+    :rtype: str
+    """
+    filename = blob_name.split("/")[-1]
+    root_name, _ = filename.rsplit(".", 1)
```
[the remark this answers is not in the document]
spencer-skylight: I have tripped over this many times - in various tools and libraries I've used, sometimes filename includes extension, and other times not.  Naming for files is all over the place.

Straw man to standardize on:
`file_location` = /mydirectory/myfile.ext
`file_path` = /mydirectory/
`filename` = myfile
`extension` = ext
`full_filename` = myfile.ext

In this case we have a before and after filename, so maybe we could rename `root_name` to `original_filename`.

spencer-skylight (changes requested):
Partway through my review - will finish up next week!

celestemespinoza:
Task linked: [CU-29av44b Refactor Code that Doesn't Fit Design Principles](https://app.clickup.com/t/29av44b)

src/FunctionApps/python/tests/IntakePipeline/test_pipeline.py
```diff
@@ -35,10 +35,10 @@ def partial_failure_message():
...
 
 
-@mock.patch("IntakePipeline.standardize_patient_name")
-@mock.patch("IntakePipeline.standardize_patient_phone")
-@mock.patch("IntakePipeline.geocode_patient_address")
-@mock.patch("IntakePipeline.add_patient_identifier")
+@mock.patch("IntakePipeline.standardize_patient_names_in_bundle")
+@mock.patch("IntakePipeline.standardize_all_phones_in_bundle")
+@mock.patch("IntakePipeline.geocode_patients_in_bundle")
+@mock.patch("IntakePipeline.add_linking_identifier_to_patients_in_bundle")
```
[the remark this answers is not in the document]
nickclyde: Is `_in_bundle` necessary, since a bundle is one of the arguments, isn't it kind of implied?

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -4,83 +4,54 @@
...
+
+def get_fhirserver_cred_manager(fhir_url: str) -> AzureFhirserverCredentialManager:
+    """
+    Get an instance of the Azure FHIR Server credential manager configured
+    for a given FHIR server url.
+    :param str fhir_url: The url of the FHIR server to access
+    :return: The credentials manager
+    :rtype: AzureFhirserverCredentialManager
+    """
     return AzureFhirserverCredentialManager(fhir_url)
```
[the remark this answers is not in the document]
nickclyde: I agree with Bryan - no need to create wrapper functions like this that just create and return class instances. If someone is reading this code, they almost certainly know how to create class instances

src/lib/phdi-building-blocks/phdi_building_blocks/classes/geocode_result.py
```diff
@@ -0,0 +1,20 @@
...
+
+    address: List[str]
+    city: str
+    state: str
+    zipcode: str
+    county_fips: str
+    county_name: str
+    lat: float
+    lng: float
+    precision: str
```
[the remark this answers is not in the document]
nickclyde: Agreed - only classes that are both very long and used in multiple files should get their own file, but even then they don't need to be in their own `classes` folder.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -89,6 +60,8 @@ def upload_bundle_to_fhir_server(bundle: dict, access_token: str, fhir_url: str)
     adapter = HTTPAdapter(max_retries=retry_strategy)
     http = requests.Session()
     http.mount("https://", adapter)
+
+    # Now, actually try to complete the API request
```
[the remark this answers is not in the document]
nickclyde: This makes sense to me. I'm willing to give it a shot!

src/lib/phdi-building-blocks/phdi_building_blocks/geo.py
```diff
@@ -1,119 +1,193 @@
...
+            state=smartystreets_result.components.state_abbreviation,
+            zipcode=smartystreets_result.components.zipcode,
+            county_fips=smartystreets_result.metadata.county_fips,
+            county_name=smartystreets_result.metadata.county_name,
+            lat=smartystreets_result.metadata.latitude,
+            lng=smartystreets_result.metadata.longitude,
+            precision=smartystreets_result.metadata.precision,
         )
 
+    return
```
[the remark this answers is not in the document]
nickclyde: Coming from Javascript land, I am very used to functions potentially returning a null. But I would definitely agree that a function should never return two different non-null types.

src/lib/phdi-building-blocks/phdi_building_blocks/geo.py
```diff
@@ -1,119 +1,193 @@
...
+    """
+
+    raw_one_line = get_one_line_address(address)
+    geocoded_result = get_geocoder_result(client, raw_one_line)
+
+    # Determine if standardized result is different from raw data
+    # Default assumption is addresses are same, in case there's no
+    # valid geocoder result
+    address_is_different = False
+    if geocoded_result:
```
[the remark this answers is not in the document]
nickclyde: Personally I'm a fan of relying on truthiness, but again, that's just me being a JS dev 😅 I don't really mind either way. Being explicit does avoid weird gotchas like if you actually want to handle some falsey values, but I don't think that's the case here.

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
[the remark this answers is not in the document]
spencer-skylight: Yeah, we should probably omit the word "standardized" here because there is no guarantee they actually have been standardized.

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
```diff
@@ -1,68 +1,100 @@
...
-                address = next(
-                    (addr for addr in patient["address"] if addr.get("use") == "home"),
-                    patient["address"][0],
-                )
-                address_line = " ".join(address.get("line", []))
-                address_line += f" {address.get('city')}, {address.get('state')}"
-                if "postalCode" in address and address["postalCode"]:
-                    address_line += f" {address['postalCode']}"
+        # Generate and store unique hash code
+        link_str = name_str + "-" + patient["birthDate"] + "-" + address_line
```
spencer-skylight, line 45: Probably out of scope for this PR, but should we be enforcing some minimum quality criteria?  For example:
* Last name with at least 2 characters
* Valid birthdate
* Address line

There is a serious possibility if birthdate and address are missing, we could assign the the same ID to different patients.

Another edge case to worry about - hospitals often assign trauma names to patients who come in to the ER.  This could mean we get many patients with the same name (e.g. John Doe) and DOB.  Also these names will change to their actual name later in the visit and could trigger mismatches.

If others think this is worth recording, I can write up tickets.

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
```diff
@@ -1,68 +1,100 @@
...
+    :param str field: The field to extract
+    :param str use: The use the field must have to qualify
+    :param int default_field: The index of the field type to treat as
+    the default return type if no field with the requested use case is
+    found
+    :return: The requested use-case-type field
+    :rtype: str
+    """
+    return next(
+        (item for item in patient[field] if item.get("use") == use),
```
spencer-skylight, line 82: This assumes that patient[field] exists as well as patient[field][default_field] else I think it will throw a KeyError.  I don't think this is ideal?  If we decide to use this code as-is, we should document this behavior better.

src/FunctionApps/python/FhirServerExport/__init__.py
```diff
@@ -8,19 +8,27 @@
 
 
 def main(req: func.HttpRequest) -> func.HttpResponse:
+    """
+    Main function that activates the FHIR export function app utility.
+    An HTTP request is used to kick off the procedure. For more information,
+    see the README file accompanying the FhirServerExport function app.
+    :param func.HttpRequest req: The request initiating a FHIR export.
+    :return: The response from the FHIR server
+    :rtype: func.HttpResponse
```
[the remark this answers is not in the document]
nickclyde: Actually I misunderstood. Sphinx definitely pulls the types from the function signature, so it's not necessary to document the types in our comments. However, the [config option](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_typehints) `autodoc_typehints` allows us to configure whether the docs show the types of parameters in the function signature, in the parameter descriptions, or both. Does anyone have a preference? The default is in the signature, only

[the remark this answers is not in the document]
spencer-skylight: My vote is signature only - I can't think of a reason we'd need to double-document.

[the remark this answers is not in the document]
bryan-skylight: My vote is for parameters only. I think it makes the signature cleaner and moves the types to be next to where the description of the parameter is. Happy to go with majority rule here though. FWIW, pandas places the type with the parameter ([source](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.groupby.html)).

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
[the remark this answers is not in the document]
bryan-skylight: I really like this. But your last sentence doesn't seem to vibe with your recommendation. Wouldn't we change `filename` to be `full_filename`, `root_name` to `filename`, and `_` to `ext`? Or am I misunderstanding?

src/FunctionApps/python/FhirServerExport/__init__.py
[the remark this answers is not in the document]
DanPaseltiner: I concur have worked on the phone number PR and now the schema code since we started documenting like this I'm getting tired of so much redundant documentation. So if our doc generation tool doesn't need it I would like to stop.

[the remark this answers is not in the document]
nickclyde: I just looked w/ Bryan. We confirmed that Sphinx can get both the parameter and return types from the function signature, so there's no need to include them in the docstring. We also looked at both config options for the output and agreed that it looks nicer to show the types next to the parameter descriptions rather than in the function signature, so I'm going to set that config option as long as there's no one against! And I will go through and remove the redundant types from the docstrings.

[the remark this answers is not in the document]
spencer-skylight: Sorry I misunderstood the question and thought we were asking where to document type _in our code_.

I don't have a strong preference where to tell Sphinx to put it in the html output, but would lean towards parameters only as Bryan suggested.

src/FunctionApps/python/requirements.txt
```diff
@@ -4,7 +4,7 @@ azure-functions
 azure-identity
 azure-storage-blob
 hl7
-phdi_building_blocks @ git+https://github.com/CDCgov/prime-public-health-data-infrastructure@main#subdirectory=src/lib/phdi-building-blocks
+phdi_building_blocks @ git+https://github.com/CDCgov/prime-public-health-data-infrastructure@principle-refactor#subdirectory=src/lib/phdi-building-blocks
```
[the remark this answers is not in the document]
nickclyde: I got this working! [See here](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/125/files#diff-245392b692a50c38ecab4381b118862db514035c10983f3bd4f4b7f1f4be4692R62)

.github/workflows/test.yaml
```diff
@@ -59,6 +59,7 @@ jobs:
         working-directory: ./src/FunctionApps/${{matrix.function-to-test}}
         run: |
           pip install -U pip
+          sed -i "\|prime-public-health-data-infrastructure@|s|main|$GITHUB_HEAD_REF|" requirements.txt
```
spencer-skylight, line 62: :guitar:

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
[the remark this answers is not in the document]
DanPaseltiner: I like the idea of agreeing on a standard file and path names. @spencer-skylight is spot on that this can be all over the place. I don't have strong feelings on the exact names we choose, but would like to ask that whatever the decision is be recorded in some kind of phdi_build_blocks style guide. I know we aren't looking to flesh out a whole guide currently, but what about a place holder .md file with one section on file and path variable naming? I'd hate for this to get lost in one conversation buried in a reasonably large closed and merged PR.

src/lib/phdi-building-blocks/phdi_building_blocks/azure_blob.py
```diff
@@ -5,7 +5,16 @@
```
bryan-skylight, line 5: I didn't know where else to put this, but it just dawned on me that this module is too specific. I'd like to see this broadened out to include all of our Azure functionality (for now; we can refactor later if necessary). This would include generating credentials, and anything specific to Azure's implementation of the FHIR server. `fhir.py` would then hold code that is generalized to the FHIR API itself.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -14,50 +14,45 @@
...
-def generate_filename(blob_name: str, message_index: int) -> str:
-    """Strip the file type suffix from the blob name, and append message index."""
-    fname = blob_name.split("/")[-1]
-    fname, _ = fname.rsplit(".", 1)
-    return f"{fname}-{message_index}"
-
-
-class AzureFhirserverCredentialManager:
-    """Manager for handling Azure credentials for access to the FHIR server"""
+class AzureFhirServerCredentialManager:
```
bryan-skylight, line 17: This whole class should be moved to what is currently named `azure_blob.py` and the file should be renamed to `azure.py`

src/lib/phdi-building-blocks/phdi_building_blocks/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "0.1.0"
+__version__ = "0.1.1"
```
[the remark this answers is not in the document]
bamader: I'm in favor of adopting Semantic Versioning (or SemVer: https://semver.org/; more friendly link https://www.geeksforgeeks.org/introduction-semantic-versioning/). Basically, leftmost number is for major releases that are such new functions they require API or invocation changes. Middle number is minor releases that introduce new functionality without breaking the API or invocation signature. Right most number are simple patches that fix bugs. By this standard I actually goofed and should've upgraded us to version 1.1.0 (as I'd argue this work changes the ways the functions are called).

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
```diff
@@ -1,68 +1,100 @@
...
+            patient["identifier"] = []
 
-            if "identifier" not in patient:
-                patient["identifier"] = []
+        patient["identifier"].append(
+            {
+                "value": hashcode,
+                # Note: this system value corresponds to the FHIR specification
+                # for a globally used / generated ID or UUID--the standard here
+                # is to make the use "temporary" even if it's not
```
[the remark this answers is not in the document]
bamader: Spencer, I think you're right. Here's the ID registry page that the specific reference code comes from: https://build.fhir.org/identifier-registry.html. When I was in my much younger PHDI days and implemented this temp storage, I was still struggling to read FHIR docs. I think you're correct that I misinterpreted what constituted a URI and what the system parameter should be used for. I'm open to using the URI you created in your comment, if others are!

src/lib/phdi-building-blocks/phdi_building_blocks/__init__.py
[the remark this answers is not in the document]
bamader: Ah you are correct and I have played myself. I'll leave it as is for now and we can write a ticket to put up an ADR around it (which I just threw into the document)

src/lib/phdi-building-blocks/phdi_building_blocks/classes/creds_manager.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        return self.fhir_url
+
+    def get_access_token(self, token_reuse_tolerance: float = 10.0) -> AccessToken:
+        """
+        Obtain an access token for the FHIR server the manager is pointed at.
+        If the token is already set for this object and is not about to expire
+        (within token_reuse_tolerance parameter), then return the existing token.
+        Otherwise, request a new one.
+        :param float token_reuse_tolerance: Number of seconds before expiration; it
```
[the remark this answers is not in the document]
bamader: I'm going to de-prioritize editing this parameter since Spencer has flagged it for further work in a separate PR

src/lib/phdi-building-blocks/phdi_building_blocks/conversion.py
```diff
@@ -72,75 +78,102 @@ def clean_message(message: str) -> str:
...
     Each group of integers is truncated to conform to the HL7 specification:
-    first integer group: max 14 digits
-    following decimal point: max 4 digits
-    following +/- (timezone): 4 digits
+      - first integer group: max 14 digits
+      - following decimal point: max 4 digits
+      - following +/- (timezone): 4 digits
+
+    This normalization facilitates downstream processing using, e.g.
+    Azure, which has particular requirements for dates.
```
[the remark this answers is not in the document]
bamader: Lol CTRL + F on "Azure" incoming

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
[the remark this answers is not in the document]
bamader: I think we should definitely put this in both a style guide (eventually) and potentially in an ADR in the mean time. That could actually go along well with the ADR around versioning and the adoption of SemVer. In this case, root was intended to be the file without the extension, which it looks like was grokked correctly, but I agree the naming for these kinds of things is terrible. I don't like how the proposed system uses `file_path` to refer to the parent directory when to me, a file path intuitively refers to the actual full path to the file (i.e. corresponds to file_location), but that's getting into semantics. For now, I'll change the variables to be super descriptive versions of themselves until we figure out what semantics we'd like to go with.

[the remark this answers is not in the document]
DanPaseltiner: I realize this is huge PR that has been in review for a long time, and I really don't want to cause further delay, but I'm going to push back on punting setting a standard around file names and paths. To my mind this is exactly the sort of thing that should come out of this PR in order to prevent the need for future style-motivated refactors. In PR 129, that I'm getting ready to merge, does a bunch of file I/O. Several times during development  I came back to this conversation to see if it had resulted in a decision so that I could adopt it and prevent the need for future refactoring. The longer we go without an answer around this the more variable names we will have to clean up later. To this end what do you y'all this of the following:

full_file_path = /mydirectory/myfile.ext
parent_directory_path = /mydirectory/
filename = myfile
extension = ext
full_extension = .ext
full_filename = myfile.ext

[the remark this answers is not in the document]
bryan-skylight: +1 to not punting. This is definitely the type of stuff that should come out of this PR.

Given the two proposed taxonomy's, I'd prefer we go with Spencer's. I think it's simpler in its naming, and I can't foresee a scenario where needing to classify something as the `parent_path` compared to just the `file_path` would be a necessary distinction. I also don't think we need to differentiate between the full extension, and the non-full extension. Instead, we should just always assume that the extension does _not_ include the period.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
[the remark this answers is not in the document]
bamader: This is now implemented 👍

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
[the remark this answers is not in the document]
DanPaseltiner: Personally, I'm perfectly happy with Spencer's nomenclature. I was just trying to address @bamader's comment about `file_path`, to move towards consensus. I included `extension` and `full_extension` because disambiguating between with and without the period isn't something that had occurred to me before. Provided we agree to always go without the period I don't see much value in it.

bamader:
Files for Brandon:
- fhir.py
- conversion.py
- IntakePipeline/init.py
- azure_blob.py (to be renamed and consolidated with a class)
- linkage.py
- utils.py
Files for Bryan:
- test_pipeline.py
- geo.py
- standardize.py
- test_geo.py (Dan left comment about a small helper function or two missing test coverage)
- test_standardize.py (same deal as above)

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -204,10 +206,19 @@ def _compose_export_url(
...
+    continues. If the response is 200, then the export files are ready, and we
+    return the HTTP response. Any other status either indicates an error or
+    unexpected condition. In this case raise an error.
+    :param str poll_url: The endpoint the FHIR server gave us to query for if
+    our files are ready
+    :param str access_token: The access token we use to authenticate with the
+    FHIR server
+    :return: None if the files didn't get processed in time, OR a response with
+    the relevant status codes and exported records if the files did get processed
+    :rtype: Either None, or a Response from Requests
```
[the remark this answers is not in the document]
bamader: Looks like whatever this comment is referring to no longer exists in the code. It _might_ be referring to the helper function on lines 182-214 that checks polling status, which is why I'm going to leave it unresolved even though github marked it as outdated. Given what Spencer has said about this lining up with the `polling` module's behavior, I would argue we should not write another helper function to standardize the return type and should instead leave the code on 182-214 as is. The user should never be calling this function directly since it's only used implicitly by the actual `export_from_fhir_server_poll` on line 236, which means they shouldn't be confused about the return types. Conversely, adding yet another function to this function chain would seem to me to make working with the `polling` module more difficult, and create the same problem you had with the `geo` module, Bryan, where there were too many helper functions getting called in a chain.

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -49,6 +61,7 @@ def run_pipeline(
     valid_output_path = get_required_config("VALID_OUTPUT_CONTAINER_PATH")
     invalid_output_path = get_required_config("INVALID_OUTPUT_CONTAINER_PATH")
```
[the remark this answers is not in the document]
bamader: Although this means we'll have to change the values in Azure as well, so actually I'm going to hold off on this until I confer with Bryan.

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
bamader: From what we've seen, yeah given name is always in a list (not only does given include middle name, but it's list structured for people with hyphenated or multiple first names). I've added a default empty string to the `get` for family name, so that if no name is provided we wind up with `[""]` as `name_parts`. When we later call the `join` method to build the name string, this will lead to the final value being an empty string, as is desired.

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
bamader: Bryan, yeah "use" is actually a word with a specific FHIR-related meaning. I've tried to expound on this a little bit more in the documentation for the function, but I've also implemented the requested generalizing nomenclature changes.

src/lib/phdi-building-blocks/phdi_building_blocks/classes/creds_manager.py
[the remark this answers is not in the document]
bryan-skylight: I agree with holding off on this for this PR and creating a new ticket for it.

.github/workflows/test.yaml
bryan-skylight: This is interesting because I trying it using Git Bash on Windows and it doesn't appear to be changing anything. But I trust that this is working, and we'll find out if it doesn't come test time. 😅

src/lib/phdi-building-blocks/phdi_building_blocks/geo.py
```diff
@@ -1,119 +1,193 @@
...
     client.send_lookup(lookup)
 
+    # Valid responses have results with lat/long
     if lookup.result and lookup.result[0].metadata.latitude:
-        res = lookup.result[0]
-        addr = [res.delivery_line_1]
-        if res.delivery_line_2:
-            addr.append(res.delivery_line_2)
+        smartystreets_result = lookup.result[0]
+        coded_address = [smartystreets_result.delivery_line_1]
```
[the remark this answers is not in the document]
bryan-skylight: After talking with Brandon, we're going to change this to `street_address` to better represent what it is.

src/lib/phdi-building-blocks/phdi_building_blocks/geo.py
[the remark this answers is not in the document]
bryan-skylight: We all seem to agree that returning `None` in addition to a single object type is fine, so I'm happy to close this out as resolved.

.github/workflows/test.yaml
bryan-skylight: Yup, I see my confusion now. I misinterpreted the `prime-public-health-data-infrastructure@` to be the string being replaced, as opposed to a preceding string to help identify where `main` should be replaced with `$GITHUB_HEAD_REF`. This all makes sense now.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
bryan-skylight: See comments regarding this functionality in the `geo.py` file. Removing here for the same reasons.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
bryan-skylight: I've opted to implement an approach that I think works more generally. I tried searching for a Python module that would do what we want, and I think the most robust option would be `NLTK`, but that's a pretty big module and seems like overkill for our needs. A simple homebrewed solution should suffice for now. We can re-evaluate later as our needs grow larger.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
bryan-skylight: I've implemented this change such that we no longer provide a named transformation, but instead allow the user to customize which transformations they'd like to apply where applicable. I believe this makes the code much more extensible and reusable, while also reducing the amount of information a user must have about our code base in order to use it.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
bryan-skylight: This is now implemented.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
bryan-skylight: We're not currently using this functionality in the code, and I can't foresee an instance where we would. I'm going to delete this for now, and should we find that we need it later on we can resurrect it.

src/FunctionApps/python/IntakePipeline/__init__.py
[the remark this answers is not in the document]
bamader: After talking with Bryan, I think we're going to leave these as is for now. If we run into a problem with multiple variable names conflicting down the road we'll add in the prefixing then, but right now this is just a hair outside the scope of the PR and we haven't actually yet run into this problem, so no sense putting the cart before the horse.

src/FunctionApps/python/tests/IntakePipeline/test_pipeline.py
```diff
@@ -78,18 +87,16 @@ def test_pipeline_valid_message(
     patched_name_standardization.assert_called_with(
         {"resourceType": "Bundle", "entry": [{"hello": "world"}]}
     )
-    patched_phone_standardization.assert_called_with(
-        {"resourceType": "Bundle", "entry": [{"hello": "world"}]}
-    )
+    patched_phone_standardization.assert_called_with(patched_standardized_name_data)
```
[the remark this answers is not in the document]
bamader: No, I think this is correct as written. If you look at lines 62 to 72, you'll see that the objects named `X_data` are the return values of the transform that standardizes `X`. So, `patched_phone_standardization` as a mock will generate the returned data `standardized_phone_data`. The actual phone standardization itself is run on the _name_ standardized data, which means this assertion is correct because `patched_phone_standardization` _should_ be called on the result of `patched_name_standardiation` and not on its own resulting data.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
bryan-skylight: I've now decided that this is slightly out of scope for this PR. This isn't related to human-centered design, per se, and is more related to generalization and scaling. Going to mark as resolved.

src/lib/phdi-building-blocks/tests/test_standardize.py
bryan-skylight: I agree with this, but at this point I think we should consider it outside the scope of this PR, which focuses on human-centered design changes. I'll create a series of tickets that represent expanding test coverage in each of our modules.

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
bryan-skylight: This is probably outside the scope of this PR, but is something we should do. I've added a TODO, and a fast follow to this PR should be to turn all of the TODOs into tickets.

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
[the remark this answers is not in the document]
bryan-skylight: I've added a TODO for tihs and we can draft up a ticket to address it.

src/lib/phdi-building-blocks/phdi_building_blocks/__init__.py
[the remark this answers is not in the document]
bryan-skylight: I've written [a ticket](https://app.clickup.com/t/2xqtzqy) to reflect this work, so I'm going to close out this conversation.

src/lib/phdi-building-blocks/tests/test_geo.py
bryan-skylight: I'm going to generate a series of tickets around ensuring our test coverage is as good as it can be for each module and function. While I wholeheartedly agree with what you're saying, Dan, this PR has been open for too long and is addressing too many things. This was an error on my part, and now I'm just trying to nip it in the bud, get it landed, and adjust for mistakes with some fast follow tickets, such as those for extending our tests to explicitly test newly created functions.

src/lib/phdi-building-blocks/tests/test_linkage.py
bryan-skylight: See prior comments regarding our handling of necessary tests.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
bryan-skylight: No longer an issue.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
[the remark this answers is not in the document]
bryan-skylight: The consensus is to go with Spencer's recommendation, which has been added to a draft version of a Design Guide. Thanks for the great back and forth here!

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -139,6 +121,8 @@ def export_from_fhir_server(
         container=container,
     )
     logging.debug(f"Composed export URL: {export_url}")
+
+    # Open connection to the export operation and kickoff process
```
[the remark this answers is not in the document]
bryan-skylight: This is now implemented.

src/lib/phdi-building-blocks/phdi_building_blocks/geo.py
```diff
@@ -1,119 +1,193 @@
...
         ClientBuilder(creds)
         .with_licenses(["us-standard-cloud"])
         .build_us_street_api_client()
     )
 
 
-def geocode_patient_address(bundle: dict, client: us_street.Client) -> dict:
-    """Given a FHIR bundle and a SmartyStreets client, geocode all patient addresses
-    in all patient resources in the bundle."""
+def geocode_patients_in_bundle(
```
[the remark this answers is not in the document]
bryan-skylight: The `_in_bundle` suffix has been removed, and TODOs have been added to find ways to generalize functions that refer to a specific resource.

src/lib/phdi-building-blocks/phdi_building_blocks/geo.py
bryan-skylight: After speaking with Brandon, I believe we've cleaned this up a bit and agreed that it's fine.

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -59,31 +72,40 @@ def run_pipeline(
...
             logging.warning(
                 "Attempted to store preexisting resource: "
                 + f"{message_mappings['filename']}.fhir"
             )
 
-        upload_bundle_to_fhir_server(bundle, access_token, fhir_url)
+        # Don't forget to import the bundle to the FHIR server as well
+        upload_bundle_to_fhir_server(standardized_bundle, access_token, fhir_url)
+
+    # For some reason, the HL7 message failed to convert
```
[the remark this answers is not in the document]
bryan-skylight: This has been addressed.

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -120,9 +142,12 @@ def main(blob: func.InputStream) -> None:
     It is responsible for splitting an incoming batch file (or individual message)
     into a list of individual messages.  Each individual message is passed to the
     processing pipeline.
+    :param func.InputStream blob: The blob that's about to go on
+    the journey of a lifetime
```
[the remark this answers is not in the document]
bryan-skylight: I've modified this comment to be more descriptive.

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -133,10 +158,13 @@ def main(blob: func.InputStream) -> None:
         messages = convert_batch_messages_to_list(
             blob.read().decode("utf-8", errors="ignore")
         )
-        message_mappings = get_file_type_mappings(blob.name)
 
+        # Once we have the file type mappings, run through all
+        # the blobs and send them down the pipeline
```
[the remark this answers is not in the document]
bryan-skylight: This has been implemented.

src/lib/phdi-building-blocks/phdi_building_blocks/linkage.py
bryan-skylight: This will be added as a TODO, which will get turned into a ticket quickly after this PR lands.

src/lib/phdi-building-blocks/phdi_building_blocks/azure_blob.py
bryan-skylight: This has been converted to `azure.py`. Thanks, team!

bryan-skylight (approved):

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
```diff
@@ -1,93 +1,99 @@
...
-        for address in resource.get("resource").get("address"):
-            country = address.get("country")
-            countries.append(standardize_country(country, code_type))
+    for address in patient.get("address"):
+        country = address.get("country")
+        countries.append(standardize_country(country, code_type))
     return countries
 
 
-def standardize_phone(raw: str, countries: List = [None, "US"]) -> str:
```
DanPaseltiner, line 90: I disagree with the choice to rename this function. as it does more than just standardize a phone number's country. The call on line 131 puts the entire phone number in the E164 international standard. In fact the country standardization is partly optional if the user use the default value of `countries`

src/lib/phdi-building-blocks/phdi_building_blocks/utils.py
```diff
@@ -1,10 +1,175 @@
...
+            return
+    elif request_type == "GET":
+        try:
+            response = requests.get(
+                url,
+                headers=headers,
+            )
+            return response
+        except Exception:
+            logging.exception("GET request to " + url + " failed.")
```
DanPaseltiner, line 175: I know this will be one more conversation that has to be resolved, but I just want to say that I think this work here is awesome!

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
bamader: After looking at the code and how we're using this function, I agree strongly with Dan. The function was written for the most general usage our particular case requires--we can already extract countries from any object. Bryan, where I think we run into a problem with the potential implementation you want to see in the future (i.e. passing in a bundle and using `find_resource_of_type`) is that that will return a list of _all_ resources of the given type in the bundle, but we want to build up a list of countries for a single _individual_ resource. I.e. we don't want the returned list to have all countries that occur across all addresses in the bundle, but rather just the countries that occur for one specific resource. Consider a bundle with two patients, X and Y, each with two addresses: X has country USA and country UK, and Y has country USA and country CA. The returned list would have three elements, which would then cause problems on line 210 when we call `phone_country_standardiation` for each individual patient and there are only two relevant elements in each patient list. I'm going to leave the existing implementation as it is with only a revised name for clarity, because it seems the most general use that balances between both of your desires, but we can open a TODO or fast follows if we really want to change it.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
bamader: After looking at the code again, I think I'm going to mark this one as resolved. There doesn't seem to be a whole lot of semantic difference between the two function names, and while it's true that `standardize_phone` is perhaps more descriptive of the resulting international standard result, the actual usage difference there is pretty negligible given that standardization into the international format will likely come about at least in part due to a country code.

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
bryan-skylight: That makes sense to me. The solution then is to revert this back to how Dan wrote it originally, right?

DanPaseltiner (approved):
Thanks for all your work on this @bamader, @bryan-skylight, and @nickclyde!

spencer-skylight (approved):
LGTM

src/lib/phdi-building-blocks/phdi_building_blocks/standardize.py
bryan-skylight: This function has been reverted, so all is well.

PR #129, opened 2022-06-06
Dan/initial fhir to parquet

DanPaseltiner:
This PR adds the following functionality:

1. Querying a FHIR server.
2. Reading a YAML schema config file (example in `examples/datamart-from-FHIR-demo/example_schema.yaml`)
3. Combing 1 and 2 to query a FHIR server for the resource types required for a schema, extract the specific fields necessary and write them in tabular form to parquet.

LIMITATIONS:

1. The approach to querying the FHIR to generate schemas is basic. Currently we are only able to request all the resources of a given type. Improving on this is definitely an area for future work.
2. This work only supports schemas containing tables of values from a single resource type. ~~Generating tables that contain fields from a variety of resources types will need to be implemented in the future.~~ Currently sub-directories for each table in a schema are created and resource type specific tables are written to them. This allows this work to process any schema a user might create will not yet dealing with the problem of combining these resource type specific tables, or executing more sophisticated searches.

`FHIR-parquet-explore/read_write_demo.py` is an example of putting all of this together. It should probably move to some other location or be removed from the repo, but I thought it might be helpful for review of the rest of the code and is probably pretty close to we'll for the 6/30 VA demo.

A note on FHIR path:
Initially this work relied on code we developed to parse FHIR resources in a fairly naive homebrewed manner. Thanks to some conversations with @spencer-skylight it turns out that HL7 has released a standard for parsing FHIR resources/bundles called [FHIRPath](https://hl7.org/fhirpath). Given the robustness of the standard we feel that relying on it for parsing instead of our own tooling is the right choice. There are a variety of implementation of the FHIRPath standard available, for the time being this work uses [fhirpathpy ](https://github.com/beda-software/fhirpath-py) however a larger conversation about precisely which FHIRPath implementation we want to move forward with is probably necessary.

TO DO:

1. ~~Write unit testing.~~
2. ~~Tweak doc strings based on recent conversations.~~


celestemespinoza:
Task linked: [CU-2vdhe5d Implement Fetchers and Parsers in Schema Application](https://app.clickup.com/t/2vdhe5d)

celestemespinoza:
Task linked: [CU-2y7qfm2 Save data exported from FHIR Server to persistent data store](https://app.clickup.com/t/2y7qfm2)

celestemespinoza:
Task linked: [CU-2f6wwuq FHIR Tabular Export: Flatten and Explode FHIR data](https://app.clickup.com/t/2f6wwuq)

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+    resource_schema = schema.get(resource.get("resourceType", ""))
+    if resource_schema is None:
+        return data
+    for field in resource_schema.keys():
+
+        path = resource_schema[field]["fhir_path"]
+        value = evaluate(resource, path)
+
+        if len(value) == 0:
+            data[resource_schema[field]["new_name"]] = ""
```
bryan-skylight, line 89: What if the field is supposed to be a data type other than string? Probably best to set this to `None` and then handle it when we handle `include_null`.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+        else:
+            selection_criteria = resource_schema[field]["selection_criteria"]
+            value = apply_selection_criteria(value, selection_criteria)
+            data[resource_schema[field]["new_name"]] = value
+
+    return data
+
+
+def make_resource_type_table(
```
bryan-skylight, line 98: I know that we're only dealing with tables from single resources for the MVP, but I'm not sure I agree with intentionally limiting the functionality to only do that. While I agree that we maybe shouldn't be considered edge cases or more complicated cases in our code, I don't think we should be naming functions in a way that limits them to a temporary state. 

I can certainly appreciate why we'd do this, and in some instances I could even be talked into it, but I think here it's relatively straightforward to generalize this a bit to not be tied to a specific resource.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+    resource_type: str,
+    schema: dict,
+    output_path: pathlib.PosixPath,
+    output_format: Literal["parquet"],
+    credential_manager: AzureFhirserverCredentialManager,
+):
+    """
+    Given a FHIR resource type, schema, and FHIR server credential manager create a
+    table containing the field from resource type specified in the the schema.
```
bryan-skylight, line 107: This should read "containing the fields from the resource type specified in the schema"

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+
+    :param resource_type: A FHIR resource type.
+    :param schema: A schema specifying the desired values by FHIR resource type.
+    :param credential_manager: A credential manager for
+    a FHIR server.
```
bryan-skylight, line 112: Should this be on a new line?

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+    additional_page = True
+    writer = None
+    while additional_page:
+        if response.status_code != 200:
+            log_fhir_server_error(response.status_code)
+            break
+
+        # Load queried data.
+        query_result = json.loads(response.content)
+        raw_schema_data = []
```
bryan-skylight, line 130: We're overloading the word schema here. And the prefix `raw_` implies that we'll be processing it in some way, but we don't. This should likely just be named `data`.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+        make_resource_type_table(
+            resource_type, schema, output_path, output_format, credential_manager
+        )
+
+
+def write_schema_table(
+    data: List[dict],
+    output_file_name: pathlib.PosixPath,
+    file_format: Literal["parquet"],
+    writer: pq.ParquetWriter = None,
```
bryan-skylight, line 196: This function should not take in a `writer` object. It should open a connection to the file itself, long enough to write the data that it needs to, and then close the connection. The way this is currently implemented keeps the connection open longer than necessary.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+    :param schema: A schema specifying the desired values by FHIR resource type.
+    :param credential_manager: A credential manager for
+    a FHIR server.
+    """
+
+    output_path.mkdir(parents=True, exist_ok=True)
+    output_file_name = output_path / f"{resource_type}.{output_format}"
+
+    query = f"/{resource_type}"
+    response = query_fhir_server(credential_manager, query)
```
bryan-skylight, line 119: This line should be moved inside of the `while` loop, and it should take the base FHIR URL as a parameter. This will allow you to remove the redundant `response = query_fhir_server(...)` call on line 149.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+
+        # Write data to parquet
+        writer = write_schema_table(
+            raw_schema_data, output_file_name, output_format, writer
+        )
+
+        # Check for an additional page of query results.
+        for link in query_result.get("link"):
+            if link.get("relation") == "next":
+                next_page_url = link.get("url")
```
bryan-skylight, line 148: Change this variable name to `url` so that it can be reused in the `query_fhir_server(...)` call you make inside the `while` loop after the recommended change.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+                response = query_fhir_server(
```
bryan-skylight, line 149: Then we can remove this redundant call.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+                    credential_manager, specific_url=next_page_url
+                )
+                break
+            else:
+                response = None
```
bryan-skylight, line 154: Lastly, because we're not basing this on whether there was a response or not, we can change this to `additional_page = False`. and remove lines 156-157.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+
+        if response is None:
+            additional_page = False
+    if writer is not None:
```
bryan-skylight, line 158: I mention this down below, but this needs to be in the `write_schema` function.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+        writer.close()
+
+
+def generate_schema(
```
bryan-skylight, line 162: This name is counter-intuitive. You're not generating a schema, but instead are creating the data mart _from_ a schema.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+    :output_path: A path to the directory where tables of the schema should be written.
+    :output_format: Specifies the file format of the tables to be generated.
+    """
+
+    schema = load_schema(schema_path)
+    schema_name = list(schema.keys())[0]
+    schema = schema[schema_name]
+    output_path = pathlib.Path(schema_name)
+
+    credential_manager = AzureFhirserverCredentialManager(fhir_url)
```
bryan-skylight, line 184: I don't think functions should be creating their own credential managers. This is something that should be done outside of building blocks, and they should take the credential manager as an input.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+
+    :param fhir_url: URL to a FHIR server.
+    :schema_path: A path to the location of a YAML schema config file.
+    :output_path: A path to the directory where tables of the schema should be written.
+    :output_format: Specifies the file format of the tables to be generated.
+    """
+
+    schema = load_schema(schema_path)
+    schema_name = list(schema.keys())[0]
+    schema = schema[schema_name]
```
bryan-skylight, line 181: Again, this is limiting us to a single table, which won't work with the actual schema we'll be using.

```python
>>> schema = {"my_table": {"Patient": {"Identifier": {}}}, "my_other_table": {"Practitioner": {"Identifier": {}}}}
>>> schema_name = list(schema.keys())[0]
>>> schema = schema[schema_name]
>>> schema
{'Patient': {'Identifier': {}}}
```

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+def write_schema_table(
+    data: List[dict],
+    output_file_name: pathlib.PosixPath,
+    file_format: Literal["parquet"],
+    writer: pq.ParquetWriter = None,
+):
+    """
+    Write schema data to a file given the data, a path to the file including the file
```
bryan-skylight, line 199: This should read "Write data extracted from the FHIR Server to a file".

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+    name, and the file format.
+
+    :param data: A list of dictionaries specifying the data for each row of a table
+    where the keys of each dict correspond the columns and the values contain the data
```
bryan-skylight, line 203: This should read "where the keys of each dict correspond to the columns, and the values contain the data"

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
bryan-skylight: UPDATE: Just saw the comment below about not being able to append to a Parquet file. I think this is fine for now, but please add a `#TODO` comment stating that we need to figure out how to overcome this. A couple solutions I've found online require reading the existing file in, appending the new data, and writing back out. That's a lot of IO, and we'd need to consider the costs of that in the cloud, but something to consider.

I think the ideal solution would be to simply batch all of the data together and make one write operation. This will require some analysis around data sizes, but could be a viable solution.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+def get_schema_summary(
+    schema_directory: pathlib.PosixPath,
+    file_format: Literal["parquet"],
+    display_head: bool = False,
+):
+    """
+    Given a directory containing the tables comprising a schema and the appropriate file
+    extension print a summary of each table.
```
bryan-skylight, line 227: This should read "Given a directory containing tables of the specified file format, print a summary of each table."

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+
+    if file_format == "parquet":
+        table = pa.Table.from_pylist(data)
+        if writer is None:
+            writer = pq.ParquetWriter(output_file_name, table.schema)
+        writer.write_table(table=table)
+        return writer
+
+
+def get_schema_summary(
```
bryan-skylight, line 220: This should be called something like `get_metadata` or `get_summary_statistics` or `preview_tables`.

src/lib/phdi-building-blocks/pyproject.toml
```diff
@@ -14,6 +14,10 @@ azure-storage-blob = "^12.12.0"
 polling = "^0.3.2"
 phonenumbers = "^8.12.48"
 pycountry = "^22.3.5"
+PyYAML = "^6.0"
+pyarrow = "^8.0.0"
+pandas = "^1.4.2"
```
bryan-skylight, line 19: I don't recall seeing `pandas` used anywhere in here. Do we need it listed here?

bryan-skylight (changes requested):

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
spencer-skylight, line 89: If `resource_schema[field]["new_name"]` doesn't exist, would it make sense to just default to `field`?  Right now it will throw a NameError I think, which might be OK I guess if we want that to be an error scenario.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+    output_path: pathlib.Path,
+    output_format: Literal["parquet"],
+    credential_manager: AzureFhirserverCredentialManager,
+):
+    """
+    Given a FHIR resource type, schema, and FHIR server credential manager create a
+    table containing the field from resource type specified in the the schema.
+
+    :param resource_type: A FHIR resource type.
+    :param schema: A schema specifying the desired values by FHIR resource type.
```
spencer-skylight, line 110: You're missing some of the parameters in the documentation.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+    :param credential_manager: A credential manager for
```
spencer-skylight, line 111: Restating this can probably be just the token to be consistent with existing code that accesses the fhir server.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+            log_fhir_server_error(response.status_code)
+            break
+
+        # Load queried data.
+        query_result = json.loads(response.content)
+        raw_schema_data = []
+
+        # Extract values specified by schema from each resource.
+        for resource in query_result["entry"]:
+            values_from_resource = apply_schema_to_resource(
```
spencer-skylight, line 134: I think it would be nice to have a comment somewhere around here about the structure of `values_from_resource` and `raw_schema_data` - otherwise I need to jump around to figure that out.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+                next_page_url = link.get("url")
+                response = query_fhir_server(
+                    credential_manager, specific_url=next_page_url
+                )
+                break
+            else:
+                response = None
+
+        if response is None:
+            additional_page = False
```
spencer-skylight, line 157: I think you could get rid of the `additional_page` flag and just have the loop check `while response is not None` right?

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+    output_path: pathlib.Path,
+    output_format: Literal["parquet"],
+):
+    """
+    Given the url for a FHIR server, the location of a schema file, and and output
+    directory generate the specified schema and store the tables in the desired
+    location.
+
+    :param fhir_url: URL to a FHIR server.
+    :schema_path: A path to the location of a YAML schema config file.
```
spencer-skylight, line 174: prefix with `param`

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
spencer-skylight, line 98: What do you think of changing the name to `write_resource_table_by_type`?

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -295,3 +295,46 @@ def _download_export_blob(blob_url: str, encoding: str = "utf-8") -> TextIO:
...
+        full_url = specific_url + query
+
+    access_token = credential_manager.get_access_token().token
+    header = {"Authorization": f"Bearer {access_token}"}
+    response = requests.get(url=full_url, headers=header)
+
+    return response
+
+
+def log_fhir_server_error(status_code: int):
```
spencer-skylight, line 329: `-> None` (this occurs a few other places as well)

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+):
+    """
+    Write schema data to a file given the data, a path to the file including the file
+    name, and the file format.
+
+    :param data: A list of dictionaries specifying the data for each row of a table
+    where the keys of each dict correspond the columns and the values contain the data
+    for each entry in a row.
+    :param output_file_name: Full name for the file where the table is to be written.
+    :output_format: Specifies the file format of the table to be written.
```
spencer-skylight, line 206: prefix with `param`

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+    :param writer: A writer object that can be kept open between calls of this function
+    to support file formats that cannot be appended to after being written
+    (e.g. parquet).
+    """
+
+    if file_format == "parquet":
+        table = pa.Table.from_pylist(data)
+        if writer is None:
```
spencer-skylight, line 214: This is a cool paradigm :+1:

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+            writer = pq.ParquetWriter(output_file_name, table.schema)
+        writer.write_table(table=table)
+        return writer
+
+
+def get_schema_summary(
```
spencer-skylight, line 220: Seems strange to me to have a `get_` method without a return value.  Maybe change to `print_schema_summary`?

spencer-skylight (changes requested):

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -295,3 +295,46 @@ def _download_export_blob(blob_url: str, encoding: str = "utf-8") -> TextIO:
     text_buffer = io.TextIOWrapper(buffer=bytes_buffer, encoding=encoding, newline="\n")
     text_buffer.seek(0)
     return text_buffer
+
+
+def query_fhir_server(
```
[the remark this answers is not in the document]
DanPaseltiner: Completely agree.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -295,3 +295,46 @@ def _download_export_blob(blob_url: str, encoding: str = "utf-8") -> TextIO:
...
+    credential_manager: AzureFhirserverCredentialManager,
+    query: str = "",
+    specific_url: str = "",
```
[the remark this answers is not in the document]
DanPaseltiner: Changed to url for simplicity as we now just pass a url and an access token.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -295,3 +295,46 @@ def _download_export_blob(blob_url: str, encoding: str = "utf-8") -> TextIO:
...
+    # When no specific url is provided use the base url for the FHIR server from the
+    # credential manager and add the query. Otherwise use the specific url by itself.
+    if specific_url == "":
+        full_url = credential_manager.fhir_url + query
+    else:
+        full_url = specific_url + query
+
+    access_token = credential_manager.get_access_token().token
+    header = {"Authorization": f"Bearer {access_token}"}
+    response = requests.get(url=full_url, headers=header)
```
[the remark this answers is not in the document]
DanPaseltiner: I had thought about this and I'm open to it if others feel strongly. My thought was that this function was just responsible for getting a response from the FHIR server and that we should let whatever functionality it calling this function decide how to handle the response.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -295,3 +295,46 @@ def _download_export_blob(blob_url: str, encoding: str = "utf-8") -> TextIO:
...
+
+def query_fhir_server(
+    credential_manager: AzureFhirserverCredentialManager,
+    query: str = "",
+    specific_url: str = "",
+) -> requests.models.Response:
+    """
+    Given the url for a FHIR server, a query to execute via the server's API, and an
+    authentication method execute a query of the server for all resources of the
+    specified type and return the response.
```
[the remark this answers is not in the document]
DanPaseltiner: You are right. Doc string is updated, but reflects several changes to this function to simplify and generalize it.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
DanPaseltiner: Thanks!

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -295,3 +295,46 @@ def _download_export_blob(blob_url: str, encoding: str = "utf-8") -> TextIO:
...
+
+    :param status_code: Status code returned by a FHIR server
+    """
+    if status_code == 401:
+        logging.error(
+            "FHIR SERVER ERROR - Status Code 401: Failed to authenticate with the FHIR server."  # noqa
+        )
+
+    elif status_code == 404:
+        logging.error("FHIR SERVER ERROR - Status Code 404: FHIR server not found.")
```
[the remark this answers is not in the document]
DanPaseltiner: According to [HL7](https://hl7.org/fhir/http.html) the possible error codes for get requests of a FHIR server are 401, 404, 410. I added 410 and generalized the message for 404. When we add functionality to support other types of requests we can expand this function. This just started as list of errors I was encountering during development.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+    :param path: Path specifying the location of a YAML file containing a schema.
+    :return schema: A user-defined schema
+    """
+
+    with open(path, "r") as file:
+        schema = yaml.safe_load(file)
+    return schema
+
+
+def apply_selection_criteria(
```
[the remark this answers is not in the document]
DanPaseltiner: There are few reasons I chose to separate this from FHIRpath:

1. As mentioned, support for random selection which @bryan-skylight had initially requested, but has now indicated he doesn't feel as strongly about its importance. I also agree that random probably isn't super useful.
2. `fhirpathpy.evaluate` always returns a list even when something like `.first()` is specified so we will still need a line like `value = value[0]`.
3. This separation allows a user to change the selection criteria in the config file without having to touch the specified FHIRpath which might require additional technical knowledge. 

I'm happy to hear other's thoughts. I'm not opposed to doing this with FHIRpath, but did see some advantages to this approach.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -295,3 +295,38 @@ def _download_export_blob(blob_url: str, encoding: str = "utf-8") -> TextIO:
     text_buffer = io.TextIOWrapper(buffer=bytes_buffer, encoding=encoding, newline="\n")
     text_buffer.seek(0)
     return text_buffer
+
+
+def fhir_server_get(url: str, access_token: str) -> requests.models.Response:
+    """
+    Submit a get request to a FHIR server given a url and access token for
```
bryan-skylight, line 302: nit: Get should be in all caps.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+    when one is encountered during parsing.
+    """
+
+    if selection_criteria == "first":
+        value = value[0]
+    elif selection_criteria == "last":
+        value = value[-1]
+    elif selection_criteria == "random":
+        value = random.choice(value)
+    elif selection_criteria == "all":
```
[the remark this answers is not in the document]
DanPaseltiner: This will cause every condition to be checked, but I'll make the change as there aren't many conditions and evaluation is simple.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
DanPaseltiner: We know `field` because of how the loop is structured and `"new_name"` should be included in the schema config for every field. I think this is okay and that later we should create a `validate_schema` function that will ensure the user has passed a schema conforming to our required format.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
DanPaseltiner: We could also write many files containing chucks of a single table. Then read them each in turn and write to a new file. I think this would require less IO than the initial proposal, but not require ever holding all the data in memory.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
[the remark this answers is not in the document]
bryan-skylight: The fact that every condition would be checked is definitely something to keep in mind. As you've stated, there are so few here that I don't think it's an issue. 

However, I did just do some googling for best practices here and found [this SO post](https://stackoverflow.com/questions/22782785/when-to-use-if-vs-elif-in-python), which actually contradicts what I'm saying. They're stating that the use of `elif` demonstrates that the values are mutually exclusive, but several `if` statements don't. Because of this, I've changed my mind and think I was wrong. I actually think we should do it the way you're doing it. I think we can still get rid of the last `elif` statement though since it doesn't actually do anything.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
DanPaseltiner: Agreed. Just added a comment about the form of `values_from_resource`.  I think given that we initialize `raw_schema_data`, now simply `data`, as a list that we explicitly append `values_from_resource` it's form is clear. Do you agree, or would you like to see additional documentation?

FHIR-parquet-explore/read_write_demo.py
```diff
@@ -0,0 +1,20 @@
...
+    load_schema,
+    get_schema_summary,
+)
+
+if __name__ == "__main__":
+
+    # Make Schema
+    fhir_url = "https://phdi-pilot.azurehealthcareapis.com"
+    schema_path = Path("schema.yml")
+    output_path = Path("")
```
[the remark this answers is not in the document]
DanPaseltiner: I was thinking a basic schema file would be included in the same directory as the example script to make running easier for a first time user.

FHIR-parquet-explore/read_write_demo.py
```diff
@@ -0,0 +1,20 @@
...
+from phdi_building_blocks.schemas import (
+    generate_schema,
+    load_schema,
+    get_schema_summary,
+)
+
+if __name__ == "__main__":
+
+    # Make Schema
+    fhir_url = "https://phdi-pilot.azurehealthcareapis.com"
```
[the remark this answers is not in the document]
DanPaseltiner: Would you be okay just leaving this as `fhir_url = # enter the URL of your FHIR server here`?

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
DanPaseltiner: Went with @spencer-skylight's suggestions to eliminate `additional_page` and rely on `response` as I think it cleans things up further.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -0,0 +1,251 @@
...
+    by the schema.
+
+    :param resource: A FHIR resource on which to apply a schema.
+    :param schema: A schema specifying the desired values by FHIR resource type.
+    :return data: A dictionary containing the data extracted from the patient a long
+    with specified variable names.
+    """
+
+    data = {}
+    resource_schema = schema.get(resource.get("resourceType", ""))
```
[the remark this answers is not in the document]
DanPaseltiner: Agreed. I think moving forward this function could simply be passed a list of fields to extract from a resource.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
DanPaseltiner: Generalized name

DanPaseltiner: Changed name to `make_table`

examples/datamart-from-FHIR-demo/example_schema.yaml
```diff
@@ -0,0 +1,103 @@
+---
+table_1:
```
spencer-skylight, line 2: nit: at this level it's more of a collection of tables than a table right?

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -295,3 +296,38 @@ def _download_export_blob(blob_url: str, encoding: str = "utf-8") -> TextIO:
...
     text_buffer.seek(0)
     return text_buffer
+
+
+def fhir_server_get(url: str, access_token: str) -> requests.models.Response:
+    """
+    Submit a GET request to a FHIR server given a url and access token for
+    authentication.
+
+    :param url: URL specifying a get request on a FHIR server.
```
spencer-skylight, line 306: Dog piling on Bryan's comment - do you want to go all caps here for GET?

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -295,3 +296,38 @@ def _download_export_blob(blob_url: str, encoding: str = "utf-8") -> TextIO:
...
+    response = requests.get(url=url, headers=header)
+    log_fhir_server_error(response.status_code)
+
+    return response
+
+
+def log_fhir_server_error(status_code: int) -> None:
+    """Given an HTTP status code from a FHIR server's response, log the specified error.
+
+    :param status_code: Status code returned by a FHIR server
```
spencer-skylight, line 320: The only other error I can remember seeing coming back from a FHIR server is 403 - forbidden.  This happens when authentication is successful, but the user (or token) isn't authorized for the request they are trying to make.  I still think it would be better to log unexpected status code here, just by adding an else clause:
```
else:
  logging.error(f"FHIR SERVER ERROR - Status code {status_code}")
```

examples/datamart-from-FHIR-demo/example_schema.yaml
DanPaseltiner: I could be wrong, but I believe the top level is the table.

examples/datamart-from-FHIR-demo/main.py
```diff
@@ -0,0 +1,25 @@
...
+    generate_schema,
+    load_schema,
+    get_schema_summary,
+)
+
+# Set required parameters
+schema_path = "example_schema.yaml"  # Path to a schema config file. Included example used by default.
+output_path = "."  # Path to directory where files will be written
+output_format = "parquet"  # File format of tables
+fhir_url = "placeholder URL"  # The URL for a FHIR server
```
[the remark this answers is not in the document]
DanPaseltiner: As an example script I think I would rather keep things as simple as possible.

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
DanPaseltiner: Makes sense. We have refactored to just pass an access token to all of these functions. We can revisit this if runtimes get long.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
DanPaseltiner: How do you feel about:
`    elif str(status_code).startswith(("4", "5")):
        logging.error(f"FHIR SERVER ERROR - Status code {status_code}")` 
        
This way if a non-error status code gets passed the function does nothing.

examples/datamart-from-FHIR-demo/example_schema.yaml
```diff
@@ -0,0 +1,103 @@
...
+      include_nulls: false
+      include_unknowns: false
+      selection_criteria: first
+      new_name: last_name
+    Phone Number:
+      fhir_path: Patient.telecom.where(system = 'phone').value
+      include_nulls: false
+      include_unknowns: false
+      selection_criteria: first
+      new_name: phone_number
```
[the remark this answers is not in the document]
DanPaseltiner: We can do that, this is intended as purely a toy schema for a new user to get up and running with something quickly. It is not the proposed schema for VA that is deliverable on 6/30.

DanPaseltiner:
As of [b7caa91](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/129/commits/b7caa91d974d0e80fe762f850c5930805ae8cad3) unit testing exists for all new functions this PR would introduce, and all tasks in the TODO section in the initial comment of this PR have been completed.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
DanPaseltiner: Addressed in [bd4eb97](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/129/commits/bd4eb97b2ef94e130456a1599c488a95a6821454) along with improvements to the quality of the unit test for `log_fhir_server_error`.

examples/datamart-from-FHIR-demo/example_schema.yaml
spencer-skylight: For the following schema:
```
table_1:
  Patient:
    Patient ID:
      fhir_path: Patient.id
      include_nulls: false
      include_unknowns: false
      selection_criteria: first
      new_name: patient_id
  Observation:
      Observation ID:
        fhir_path: Observation.id
        include_nulls: false
        include_unknowns: false
        selection_criteria: first
        new_name: observation_id
```
We'll end up with a directory structure of:
ouput_path
 |- table_1
   |- Patient.parquet
   |- Observation.parquet

I was kinda equating each parquet file as a table.  I may be misinterpreting.  I don't think this is a blocking issue, so I'm going to go ahead and approve, but may be worth getting on the same page with terminology at some point.  Thanks!

spencer-skylight (approved):

examples/datamart-from-FHIR-demo/example_schema.yaml
DanPaseltiner: Yes, this results in a `table_1` directory containing `Patient.parquet` and `Observation.parquet`. This was an intentional change I made, but didn't document well (I'll address that now). My understanding of the current schema format is that the top level represents tables within the schema. The motivation for this change is to be able to handle schemas that specify tables containing fields from multiple resources types. We haven't yet tackled exactly how we want to combine data across resources types into single tables. For the time being this allows us to process any schema a user might submit by creating resource type specific "sub-tables" for each table specified in a schema. 

From here I see a few possible paths:
1. After the sub-tables are written, read them them back in and perform joins between them to make the desired table.
2. Same as above, but keep the sub tables in memory.
3. Get more sophisticated at interpreting the schema config files and/or add additional information to them that would allow us to execute searches using more advanced features like `_include` and `_revinclude`. https://build.fhir.org/search.html#revinclude

bryan-skylight: +1 to what Dan is saying. I think for the purposes of the demo, and for MVP purposes, this is sufficient as the ECR schema only reads fields from a single resource for any given table. Moving forward, exploring the three options Dan has proposed will be a good next step.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -24,6 +24,7 @@ def generate_filename(blob_name: str, message_index: int) -> str:
 class AzureFhirserverCredentialManager:
     """Manager for handling Azure credentials for access to the FHIR server"""
 
+    # TODO: Generalize this class to decouple from Azure
```
bryan-skylight, line 27: This is _super_ nitpicky, but can we move this comment to be above the class (i.e. line 23)?

src/lib/phdi-building-blocks/tests/test_schemas.py
```diff
@@ -0,0 +1,247 @@
...
+
+    make_table(
+        schema,
+        output_path,
+        output_format,
+        fhir_url,
+        access_token,
+    )
+
+    assert len(patch_write.call_args_list) == 0
```
bryan-skylight, line 161: I'm not quite sure I understand how this assertion tests that `make_table` is working as expected when it fails. Can you elaborate on this a bit?

bryan-skylight (approved):
Looks good! Nice work on putting together a really large feature. This should satisfy the requirements for the MVP, and gives us a really nice foundation for iteration. I've left a really nitpicky comment, as well as a question about one of the tests, but I don't think either of those things are blocking.

src/lib/phdi-building-blocks/tests/test_schemas.py
DanPaseltiner: This is an overly complicated way to assert that the function wasn't called. Since writing this I learned about `assert_not_called()` and used it elsewhere in these tests. It would probably be a clearer choice here. The goal of this test was to ensure that given a `status_code != 200` we don't run any of the code in that block.

DanPaseltiner: [25fbc25](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/129/commits/25fbc2543798f8861ab233cb5f665e02eaebb030) makes the change to `assert_not_called()`.

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
DanPaseltiner: Done.

PR #131, opened 2022-06-14
Port FHIR to Tables Demo Script to Notebook

DanPaseltiner:
This PR aims to address [this ticket.](https://app.clickup.com/t/2h262rj).

A few points:

- This notebook will likely need slight modifications after the current work around addressing authentication with the FHIR server to prevent access_token expiration lands.
- The language of the tickets specifies "parquet", but I would prefer to keep things more general as our work is designed to eventually support multiple output formats and we have a ticket in the current for add support for CSVs. A new user should be able to use this notebook to generate parquet or CSV files soon.
- I'm not in love with "FHIR-to-tables" name and going to keep thinking on it. Definitely open to suggestions here.


examples/FHIR-to-tables/FHIR-to-tables.ipynb
```diff
@@ -0,0 +1,123 @@
...
+    "### Set Required Parameters"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "schema_path = \"example_schema.yaml\"  # Path to a schema config file.\n",
```
bamader, line 56: Might be worth adding that the schema file to use must be in yaml format.

examples/FHIR-to-tables/FHIR-to-tables.ipynb
```diff
@@ -0,0 +1,123 @@
...
+    "output_path = \".\"                    # Path to directory where files will be written\n",
```
bamader, line 57: We should add that the demo script creates a directory (if that's what it does, which I think is true). If it doesn't, I feel like it should and then put all the results in there.

examples/FHIR-to-tables/FHIR-to-tables.ipynb
```diff
@@ -0,0 +1,123 @@
...
+    "output_format = \"parquet\"            # File format of tables\n",
```
bamader, line 58: I would add more detail in the comment specifying that we can also support CSV outputs (or maybe I misunderstood and this isn't yet true, in which case ignore me).

DanPaseltiner: We don't yet have CSV support. There is a ticket in this sprint for it though.

examples/FHIR-to-tables/FHIR-to-tables.ipynb
DanPaseltiner: It does create a directory and we can state that explicitly.

examples/FHIR-to-tables/FHIR-to-tables.ipynb
```diff
@@ -0,0 +1,123 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Import Dependencies"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
```
spencer-skylight, line 32: It seems odd to me that we are committing the execution count as well as notebook metadata down below.  It looks like there are solutions to clean this information when committing.  Is that something we should look in to?

examples/FHIR-to-tables/FHIR-to-tables.ipynb
```diff
@@ -0,0 +1,123 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "schema_path = \"example_schema.yaml\"  # Path to a schema config file.\n",
+    "output_path = \".\"                    # Path to directory where files will be written\n",
+    "output_format = \"parquet\"            # File format of tables\n",
+    "fhir_url = \"your_fhir_url\"           # The URL for a FHIR server\n",
+    "access_token = \"your_access_token\"   # Access token for authentication with FHIR server."
```
spencer-skylight, line 60: At some point we'll need to pass in a credential manager, and use az login to sign in the user rather than supplying an access token.  We could prepare for that by adding a cell to do az login and change this code to instantiate a credential manager and obtain a token from it rather than expecting a token directly.  But, we can also wait until this becomes necessary.

spencer-skylight (approved):
A few non-blocking comments, but otherwise LGTM.

examples/FHIR-to-tables/FHIR-to-tables.ipynb
DanPaseltiner: This is a really great point. Just did a little reading on nb-clean, going to look at setting up a GitHub action to ensure notebooks are clean before allowing a merge.

examples/FHIR-to-tables/FHIR-to-tables.ipynb
DanPaseltiner: Added an additional job in the Test workflow to run [this](https://github.com/ResearchSoftwareActions/EnsureCleanNotebooksAction) git hub action that ensures notebooks have clean metadata. I have confirmed that `nb-clean` is capable of cleaning notebook metadata and that its use results in a notebook that was previously failing the new test to pass it.

examples/FHIR-to-tables/FHIR-to-tables.ipynb
DanPaseltiner: Added an additional job in the Test workflow to run [this](https://github.com/ResearchSoftwareActions/EnsureCleanNotebooksAction) git hub action that ensures notebooks have clean metadata. I have confirmed that `nb-clean` is capable of cleaning notebook metadata and that its use results in a notebook that was previously failing the new test to pass it.

examples/FHIR-to-tables/FHIR-to-tables.ipynb
DanPaseltiner: After discussing with @spencer-skylight these changes will be included in [PR 128](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/138)

examples/FHIR-to-tables/FHIR-to-tables.ipynb
DanPaseltiner: Addressed in [ed8dee6](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/131/commits/ed8dee6ad1514ac5b323122a07663390315b23ba).

examples/FHIR-to-tables/FHIR-to-tables.ipynb
DanPaseltiner: Addressed in [ed8dee6](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/131/commits/ed8dee6ad1514ac5b323122a07663390315b23ba).

celestemespinoza:
Task linked: [CU-2yzx5wu Incorporate nb-clean to Clean Metadata from Notebooks](https://app.clickup.com/t/2yzx5wu)

PR #132, opened 2022-06-16
Support failed auth retry

spencer-skylight:
### Description
Previously, calls to the FHIR server required the building block consumer to obtain an access token and pass it in to perform calls to the FHIR server.  If a token had expired, the building block calls to the FHIR server had no way to recover from an authentication error.

Now building block consumers must instead pass in a credential manager, allowing the building block to recover from an authentication error by requesting a new token and retrying the request.

More robust retry logic for other error codes (429, 500, 502, 503, 504) are also now uniformly available across supported calls to the FHIR server.


src/FunctionApps/python/FhirServerExport/__init__.py
```diff
@@ -19,11 +19,9 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
 
     cred_manager = fhir.AzureFhirserverCredentialManager(fhir_url=fhir_url)
```
bamader, line 20: Flagging this as something that will need to be changed once the refactor lands. It'll become `azure.AzureFhir...`.

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -12,6 +12,7 @@
     upload_bundle_to_fhir_server,
     generate_filename,
     get_fhirserver_cred_manager,
+    AzureFhirserverCredentialManager,
```
bamader, line 15: As above, import will now be from `azure.py`

src/FunctionApps/python/tests/FhirServerExport/test_main.py
```diff
@@ -15,9 +14,11 @@
 
 @mock.patch("FhirServerExport.fhir.download_from_export_response")
 @mock.patch("FhirServerExport.fhir.export_from_fhir_server")
-@mock.patch.object(AzureFhirserverCredentialManager, "get_access_token")
+@mock.patch("FhirServerExport.fhir.AzureFhirserverCredentialManager")
```
bamader, line 17: Another import flag (sorry if this is obnoxious, just trying to get ahead of changes I know will need to happen)

bamader (commented):
Code itself looks good to me and is a much needed improvement. I'm holding off on approval until the refactor lands, as that changes the import location of the creds manager as well as introduces the `http_request_with_retry` util that's used in a number of fhir functions (you can see I started marking those places but then realized it was endemic to the whole change). Once the refactor lands this looks good to go as long as imports get taken care of.

nickclyde:
Agreed with Brandon, the code looks great! But will probably need a little reworking after the refactor is merged, will almost certainly cause some merge conflicts. Will take another look afterwards!

src/lib/phdi-building-blocks/phdi_building_blocks/conversion.py
```diff
@@ -275,9 +282,25 @@ def convert_message_to_fhir(
             {"name": "rootTemplate", "valueString": root_template},
         ],
     }
-    response = requests.post(
-        url=url, json=data, headers={"Authorization": f"Bearer {access_token}"}
+    access_token = cred_manager.get_access_token().token
+    headers = {"Authorization": f"Bearer {access_token}"}
+
+    retry_strategy = Retry(
+        total=3,
```
KennethSkylight, line 289: I think 3 assigned as a constant. Otherwise its just a magic number

src/lib/phdi-building-blocks/phdi_building_blocks/conversion.py
```diff
@@ -275,9 +282,25 @@ def convert_message_to_fhir(
             {"name": "rootTemplate", "valueString": root_template},
         ],
     }
-    response = requests.post(
-        url=url, json=data, headers={"Authorization": f"Bearer {access_token}"}
+    access_token = cred_manager.get_access_token().token
```
KennethSkylight, line 285: It seems like this section of code is repeated in other files. Perhaps it should be moved out and imported

src/lib/phdi-building-blocks/phdi_building_blocks/fhir.py
```diff
@@ -231,12 +284,16 @@ def __export_from_fhir_server_poll_call(
 
 
 def export_from_fhir_server_poll(
-    poll_url: str, access_token: str, poll_step: float = 30, poll_timeout: float = 300
+    poll_url: str,
+    cred_manager: AzureFhirserverCredentialManager,
+    poll_step: float = 30,
```
KennethSkylight, line 289: I think these two numbers should also be constants to be clear on what the magic numbers are.

KennethSkylight (commented):
Not sure if we have a pattern in place for how to import functions that are reused in multiple files. But I think that may be necessary soon.

spencer-skylight:
Moved to new support-failed-auth-retry-hcd branch, incorporating changes from PR 125.

PR is forthcoming.

PR #133, opened 2022-06-16
Add GitHub Actions for deployment

nickclyde:
Resolves [this ticket](https://reportstream-service-request.clickup.com/t/29av3gn)

This PR creates two new GitHub Actions, one for deploying to `dev` and another for deploying to `test`. They are pretty simple, just logging into Azure/Terraform using the creds stored for CI, and running terraform apply. We will need to merge this PR before we can test running the actions, which will create the actions in [the UI here](https://github.com/CDCgov/prime-public-health-data-infrastructure/actions), then any further changes can be tested in further PRs before merging, since branches can be specified for existing actions.

### DIFF
--- a/.github/workflows/deployDev.yaml
+++ b/.github/workflows/deployDev.yaml
@@ -1,0 +1,36 @@
+name: Deploy Dev
+
+on:
+  workflow_dispatch:
+
+jobs:
+  deploy_dev:
+    name: Deploy to prime-ingestion-dev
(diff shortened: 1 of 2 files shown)


bamader (approved):
This looks good to me from my exceptionally limited understanding of how these things work (after some solid time staring lol), but I think we should wait for one more +1 here before merging to double check me.

spencer-skylight (approved):
Looks good!

PR #136, opened 2022-06-17
Use github context to get current branch for testing workflow

nickclyde:
Since $GITHUB_HEAD_REF is only available in pull requests, use the `github` [context object](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) to get the current branch instead for the testing workflow. I _think_ this should be available on the main branch as well, but we will have to merge to find out.


bryan-skylight (approved):
I give this my [flourish of approval](https://en.wikipedia.org/wiki/Flourish_of_approval)!

PR #138, opened 2022-06-20
Support failed auth retry hcd

spencer-skylight:
New branch for [PR 132 - Support failed auth retry](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/132) that incorporates merges from [PR 125 - Human Centered Design](https://github.com/CDCgov/prime-public-health-data-infrastructure/pull/132).

Comments left on PR 132 will be addressed here, but readers may follow the PR 132 link above for details.

### Description
Previously, calls to the FHIR server required the building block consumer to obtain an access token and pass it in to perform calls to the FHIR server. If a token had expired, the building block calls to the FHIR server had no way to recover from an authentication error.

Now building block consumers must instead pass in a credential manager, allowing the building block to recover from an authentication error by requesting a new token and retrying the request.

More robust retry logic for other error codes (429, 500, 502, 503, 504) are also now uniformly available across supported calls to the FHIR server.


src/FunctionApps/python/FhirServerExport/__init__.py
```diff
@@ -27,12 +27,11 @@ def main(req: func.HttpRequest) -> func.HttpResponse:
         container = ""
 
     cred_manager = AzureFhirServerCredentialManager(fhir_url=fhir_url)
```
spencer-skylight, line 29: From @bamader: Flagging this as something that will need to be changed once the refactor lands. It'll become `azure.AzureFhir...`. 

Response from @spencer-skylight: This has been addressed.

src/FunctionApps/python/IntakePipeline/__init__.py
```diff
@@ -35,7 +35,7 @@ def run_pipeline(
     message: str,
     message_mappings: Dict[str, str],
     fhir_url: str,
-    access_token: str,
+    cred_manager: AzureFhirServerCredentialManager,
```
spencer-skylight, line 38: From @bamader: As above, import will now be from azure.py

Response by @spencer-skylight: This has been addressed.

src/FunctionApps/python/tests/FhirServerExport/test_main.py
```diff
@@ -13,13 +12,12 @@
 }
 
 
-@mock.patch("FhirServerExport.fhir.download_from_export_response")
 @mock.patch("FhirServerExport.fhir.export_from_fhir_server")
-@mock.patch.object(AzureFhirServerCredentialManager, "get_access_token")
+@mock.patch("FhirServerExport.AzureFhirServerCredentialManager")
```
spencer-skylight, line 16: From @bamader : Another import flag (sorry if this is obnoxious, just trying to get ahead of changes I know will need to happen)

Response from @spencer-skylight : Addressed

spencer-skylight:
Comment from @KennethSkylight : 
> Not sure if we have a pattern in place for how to import functions that are reused in multiple files. But I think that may be necessary soon.

I think this may warrant some discussion by the team.  We are applying somewhat of a pattern where needed right now, but it may be beneficial to document the details on how this should be done and confirm we are following best practices and applying the pattern in a consistent way across our code.

spencer-skylight:
@KennethSkylight pointed out a few places where we have literal values defined directly as function defaults and other places, and suggested defining them as constants as a best practice.  

Before I make changes addressing these comments, I wanted to give the team a chance to discuss details around how this should be implemented.  As this is being discussed, we should also discuss whether related changes should be done as a part of this PR, or as a separate ticket.

src/lib/phdi-building-blocks/phdi_building_blocks/conversion.py
```diff
@@ -318,9 +324,25 @@ def convert_message_to_fhir(
...
+    adapter = HTTPAdapter(max_retries=retry_strategy)
+    http_session = requests.Session()
+    http_session.mount("https://", adapter)
+    response = http_session.post(url=url, json=data, headers=headers)
+
+    # Retry with new token in case it expired since creation (or from cache)
+    if response.status_code == 401:
+        access_token = cred_manager.get_access_token().token
+        reauth_headers = headers.copy()
+        reauth_headers["Authorization"] = f"Bearer {access_token}"
```
nickclyde, line 344: This is sort of a nit, but is there any benefit to copying the existing headers object and overwriting the token vs. just creating a new object literal?

nickclyde (approved):
Looks great! Thanks for working on this Spencer, just one non-blocking question.

src/lib/phdi-building-blocks/phdi_building_blocks/utils.py
```diff
@@ -170,10 +174,33 @@ def http_request_with_retry(
...
-            return response
         except Exception:
             logging.exception("GET request to " + url + " failed.")
+
+    if (
+        cred_manager
+        and response.status_code == 401
+        and headers.get("Authorization", "").startswith("Bearer ")
+    ):
+        new_access_token = cred_manager.get_access_token().token
```
bamader, line 189: Nice. Independent of the credential manager work, does it make sense to try to fold any other error handling into this util function? Or should we let the caller handle all of that?

src/lib/phdi-building-blocks/phdi_building_blocks/utils.py
```diff
@@ -170,10 +174,33 @@ def http_request_with_retry(
...
+
+        # New headers dict mostly for unit testing so we can test the header on the
+        # first and second attempt
+        reauth_headers = headers.copy()
+        reauth_headers["Authorization"] = f"Bearer {new_access_token}"
```
bamader, line 194: I def get how this simplifies things for unit testing, but if that's the only real reason it's here is there a way we can move this to the actual testing file rather than be held in the function invocation? I know the performance hit of maintaining one extra dict for a call is negligible, I just hate the idea that we're carrying around an extra data structure just for some test cases.

src/lib/phdi-building-blocks/phdi_building_blocks/conversion.py
spencer-skylight: This has been updated so as not to copy the headers dict, and tests just check the final call to ensure the updated access token was supplied.  

Tests also confirm there were 2 get requests.  I chatted with Bryan on this, and he agrees this is sufficient for unit tests.

PR #141, opened 2022-06-24
Ignore Notebooks not in Examples Folder

bryan-skylight:
# Description
This PR adds two lines to our `.gitignore` file, one of which ignores all Jupyter notebook files and their related checkpoints, and another line that adds an exception for notebooks listed in the `examples` directory. The idea is to allow engineers to spin up notebooks in the virtual environment we're using for development so they can prototype code and tweaks more quickly than might be possible just using VSCode.


DanPaseltiner (approved):
LGTM

bamader (approved):

PR #142, opened 2022-06-24
Default to None for Missing Values in FHIR to Data Mart ETL Pipeline

bryan-skylight:
# Description
This PR addresses an issue where non-string fields (e.g. latitude or longitude) that were missing for a resource were being defaulted to `""` instead of `None`. The issue was cashing the entire pipeline to crash when trying to write the data to a Parquet file because Parquet was trying to convert the string to a double. By setting the default value for missing fields to `None`, Parquet is able to properly handle the missing data.

# Test Plan
This has been tested in a Jupyter notebook using Patient data that included latitude and longitude. I first ran the data through the code as written, observed that it failed, and then modified the `apply_schema_to_resource` function locally to use `None` instead of `""`. I ran the data through the updated version of the code and observed that it worked.


DanPaseltiner (approved):
LGTM

PR #144, opened 2022-06-27
Demo analytics notebook

bamader:
This change set introduces a demo notebook for use with the ECR data mart to show some toy examples of how our work can:

a. pull data from cloud-hosted storage
b. read parquet files into R for analysis and manipulation
c. work through some trivial analysis and visualization problems
d. show a somewhat meaningful choropleth graph akin to the one Vincent developed months ago.

The notebook can be run through jupyter if the R Kernel is configured to do so (the notebook includes instructions on setting this up).

Tagging Bryan and Dan as the reviewers since this is client/product facing and Dan has experience with notebooks and as someone on the receiving side of work like this.

### DIFF
--- a/examples/parquet_analytics_demo.ipynb
+++ b/examples/parquet_analytics_demo.ipynb
@@ -1,0 +1,470 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "af5806d3",
+   "metadata": {},
+   "source": [
+    "# Parquet Analytics Demo Notebook"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6f5d5a85",
+   "metadata": {},
+   "source": [
+    "This notebook serves as a brief demonstration of how a set of cloud-hosted parquet files can be analyzed locally using R. While the analysis in this notebook consists of contrived / toy examples, the data being operated on is in fact _real_! It is currently stored in the FHIR server as part of the PHDI Ingestion Pipeline and has been transformed and standardized by a number of preceding building blocks."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "dc8062d0",
+   "metadata": {},
+   "source": [
+    "To run this notebook, you'll need to configure your system to recognize the R kernel. A simple tutorial can be found here: https://irkernel.github.io/installation/. To start, make sure you've installed the latest version of R, available from the CRAN repository at https://cloud.r-project.org/index.html. Then, launch R from the terminal with the `R` command. **Important:** Do not launch R from the app--the following package installation **must** be performed from the terminal (this is because the R app does not consistently respect PATH changes set in bash profiles). Run `install.packages(IRkernel)`, then run `IRkernel::installspec()`. That's it! Your notebook software can now \"see\" the R kernel on your machine."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "eca904da",
+   "metadata": {},
+   "source": [
+    "## Packages and imports"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "93cdcd4d",
+   "metadata": {},
+   "source": [
+    "We'll begin by installing the relevant analytics packages that we'll use for our examples. We'll need:\n",
+    "* `arrow`\n",
+    "* `ggplot2`\n",
+    "* `jsonlite`\n",
+    "* `geojsonio`\n",
+    "* `sf`\n",
+    "* `tidyverse`\n",
+    "\n",
+    "which you can install individually by running `install.packages(arrow)`, etc. These packages allow us to load parquet data, interact with and format data the way we want, and visualize our result. Additionally, we'll need two packages specific to accessing Microsoft's Azure platform: `AzureStor` and `AzureRMR`.\n",
+    "\n",
+    "While common practice is to use the `rgdal` package for spatial analytics purposes (see last example), `rgdal` will reach its end-of-supported lifecycle in early 2023. Hence, we recommend the new industry standard `sf` package (standing for \"simple features\")."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "f2e39b1c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Import relevant packages\n",
+    "library(arrow)\n",
+    "library(jsonlite)\n",
+    "library(AzureStor)\n",
+    "library(AzureRMR)\n",
+    "library(sf)\n",
+    "library(geojsonio)\n",
+    "library(tidyverse)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "acaf7238",
+   "metadata": {},
+   "source": [
+    "## Access and download data"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e0871e7a",
+   "metadata": {},
+   "source": [
+    "Now that we've got our package environment set up, let's load the data. This notebook demonstrates the important fact that the output of the PHDI Building Blocks pipeline are cloud-hosted files that are accessible locally for all manner of analytics. Let's access the Storage Account and pull down the parquet data to read into R."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d76d4d85",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Configure container access to Azure storage\n",
+    "az <- create_azure_login(tenant=\"9ce70869-60db-44fd-abe8-d2767077fc8f\", host=\"https://storage.azure.com/\")\n",
+    "token <- az$token$credentials$access_token\n",
+    "adls_ep <- adls_endpoint(\"https://pitestdatasa.dfs.core.windows.net/\", token=token)\n",
+    "container <- storage_container(adls_ep, \"gold\")\n",
+    "\n",
+    "# Initialize some tempfiles to buffer data into\n",
+    "phdi_conditions = tempfile()\n",
+    "phdi_organizations = tempfile()\n",
+    "phdi_compositions = tempfile()\n",
+    "phdi_patients = tempfile()\n",
+    "\n",
+    "# Perform storage download, copying data into the temp files\n",
+    "storage_download(container, \"Condition.parquet\", phdi_conditions)\n",
+    "storage_download(container, \"organization.parquet\", phdi_organizations)\n",
+    "storage_download(container, \"composition.parquet\", phdi_compositions)\n",
+    "storage_download(container, \"patient.parquet\", phdi_patients)\n",
+    "\n",
+    "# Notice that we can read-in only a subset of columns if we so choose\n",
+    "# i.e. It's not necessary to fully read the file to memory\n",
+    "conditions = read_parquet(phdi_conditions, col_select = c(\"patient_id\", \"onset_date\"))\n",
+    "organizations = read_parquet(phdi_organizations)\n",
+    "compositions = read_parquet(phdi_compositions)\n",
+    "patients = read_parquet(phdi_patients)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a83320d1",
+   "metadata": {},
+   "source": [
+    "We've now created local copies of all data that maps into the ECR schema--we have resources for Patients, Conditions, Organizations, and Compositions (between patients and organizations). Importantly, we've stored all of this data into `tempfiles` which will persist _only_ as long as our current R session. That way, when we're finished with analysis, the files will be cleared, leaving no chance for PII to linger. We'll use these file for the remainder of our analysis demo, noting that pulling files down this way allows analysts to experiment freely on data without any risk of corrupting the source of truth.\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


examples/parquet_analytics_demo.ipynb
```diff
@@ -0,0 +1,470 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d76d4d85",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Configure container access to Azure storage\n",
+    "az <- create_azure_login(tenant=\"9ce70869-60db-44fd-abe8-d2767077fc8f\", host=\"https://storage.azure.com/\")\n",
+    "token <- az$token$credentials$access_token\n",
+    "adls_ep <- adls_endpoint(\"https://pitestdatasa.dfs.core.windows.net/\", token=token)\n",
```
spencer-skylight, line 96: Keep pitestdatasa implementation-specific value out of source control.

examples/parquet_analytics_demo.ipynb
```diff
@@ -0,0 +1,470 @@
...
+    "\n",
+    "# Initialize some tempfiles to buffer data into\n",
+    "phdi_conditions = tempfile()\n",
+    "phdi_organizations = tempfile()\n",
+    "phdi_compositions = tempfile()\n",
+    "phdi_patients = tempfile()\n",
+    "\n",
+    "# Perform storage download, copying data into the temp files\n",
+    "storage_download(container, \"Condition.parquet\", phdi_conditions)\n",
+    "storage_download(container, \"organization.parquet\", phdi_organizations)\n",
```
spencer-skylight, line 107: Why is there a difference in capitalization in the parquet files?

examples/parquet_analytics_demo.ipynb
```diff
@@ -0,0 +1,470 @@
...
+   "source": [
+    "## Access and download data"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "e0871e7a",
+   "metadata": {},
+   "source": [
+    "Now that we've got our package environment set up, let's load the data. This notebook demonstrates the important fact that the output of the PHDI Building Blocks pipeline are cloud-hosted files that are accessible locally for all manner of analytics. Let's access the Storage Account and pull down the parquet data to read into R."
```
spencer-skylight, line 83: Should we mention where the data comes from (the FHIR-to-Parquet) example?

examples/parquet_analytics_demo.ipynb
```diff
@@ -0,0 +1,470 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "1d4e9a0c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "head(patients)"
```
spencer-skylight, line 136: Nit: it would be nice to put a comment here to make it painfully clear this is expected to result in an error.  That fact is already clear from the description, but it might help to have it in the comments for the tl;dr scenario.

examples/parquet_analytics_demo.ipynb
```diff
@@ -0,0 +1,470 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "2b76e6e2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Open the shapefile using the sf package\n",
+    "va_tracts = st_read(\"/Users/Brandon/Downloads/tl_2018_51_tract/tl_2018_51_tract.shp\")\n",
```
spencer-skylight, line 353: I think we should replace this location with a local variable and a dummy location

spencer-skylight (commented):
This is really cool!  It finally turns what we've been doing into interesting visual output :grin: 

I just had a few notes as someone setting up an environment to run the notebook.

examples/parquet_analytics_demo.ipynb
bamader: As above

examples/parquet_analytics_demo.ipynb
bamader: Sure, I can do that!

examples/parquet_analytics_demo.ipynb
```diff
@@ -0,0 +1,476 @@
...
+   "source": [
+    "## Packages and imports"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "93cdcd4d",
+   "metadata": {},
+   "source": [
+    "We'll begin by installing the relevant analytics packages that we'll use for our examples. We'll need:\n",
```
bryan-skylight, line 40: You need to install the Azure related libraries as well since those are required to connect to cloud storage.

examples/parquet_analytics_demo.ipynb
```diff
@@ -0,0 +1,476 @@
...
+    "# This call WILL result in an error\n",
+    "head(patients)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "0749f6fb",
+   "metadata": {},
+   "source": [
+    "This is because the `phone_number` column is nost only a nested list, but because the information stored within that column is of variable length (because some patients have phone numbers on file while others do not). Let's format this column to line up with what we'd like our data frame to be: a singular, flattened entity on which we can do joins and perform analysis."
```
bryan-skylight, line 145: typo: "nost" instead of "not"

examples/parquet_analytics_demo.ipynb
```diff
@@ -0,0 +1,476 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "e56dbcf0",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Flatten the list column to a series of regular columns\n",
+    "patients <- patients %>% unnest(phone_number)\n",
+    "patients <- subset(patients, select = -c(system, use))\n",
+    "patients <- rename(patients, phone_number = value)\n",
```
bryan-skylight, line 158: nit: This could be more "r-thonic" (that's not a thing like pythonic is, but whatever) if you changed it to:

```r
patients <- patients %>%
    unnest(phone_number) %>%
    subset(select=-c(system, use)) %>%
    rename(phone_number=value)
```

Users who are familiar with the tidyverse are likely more accustomed to seeing a sequence of data wrangling methods chained together with the pipe like this. There's nothing wrong with how you've done it, but it could increase readability for the users. Do with this information what you will.

bryan-skylight (changes requested):
Overall, this looks fantastic, and I love seeing an actual analysis being done on data that we've ingested, curated, stored, retrieved, munged again, and stored again. We've finally reached full data lifecycle here. Well done!

With that said, there are some changes I'd like to see before merging. In addition to the comments I've left, can you also add a cell towards the top of the notebook that contains just this line of code `options(repr.plot.width=10, repr.plot.height=8)` and an explanation that this sets the width and height, respectively, of the graph outputs, which will make them bigger than the default size.

examples/parquet_analytics_demo.ipynb
bryan-skylight: These file names are generated by the code we have in `main`. If this is an actual problem,  let's solve the root cause and not band-aid the symptom.

examples/parquet_analytics_demo.ipynb
```diff
@@ -0,0 +1,476 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "dc8062d0",
+   "metadata": {},
+   "source": [
+    "To run this notebook, you'll need to configure your system to recognize the R kernel. A simple tutorial can be found here: https://irkernel.github.io/installation/. To start, make sure you've installed the latest version of R, available from the CRAN repository at https://cloud.r-project.org/index.html. Then, launch R from the terminal with the `R` command. **Important:** Do not launch R from the app--the following package installation **must** be performed from the terminal (this is because the R app does not consistently respect PATH changes set in bash profiles). Run `install.packages('IRkernel')`, then run `IRkernel::installspec()`. That's it! Your notebook software can now \"see\" the R kernel on your machine."
```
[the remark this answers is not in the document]
bamader: On further reading, it doesn't appear to be. I pulled that statement from a help tutorial I used to get R working in my notebook (https://irkernel.github.io/installation/), but they don't elaborate any further. So I did some digging, and what's apparently the real issue is that the location that R installs from differs depending on whether you use CRAN (recommended), R.App / RStudio, or a package manager (namely homebrew). The frameworks and executables get put in different locations, such that if you launch from RSudio or R.App and that program isn't authorized to modify your $PATH, then changes you make to the installspec won't hold true for the root R program (the actual terminal-based interface). At least, that's what I think is happening. The recommendation then is to just install directly from CRAN (both because it's a binary installer which means no compiling from source, and because CRAN will install the libraries/frameworks/executables all in the same place which is also the first place that subsequent runs of R.App and RStudio will check when looking for the path to R) and use R within an authorized session of your terminal to make the installspec changes

TLDR: I think the advice I got in the tutorial was well-meaning but misleading. We install from CRAN and manage the installspec configuration in R terminal because it eliminates any difficulties in mucking around with $PATH configurations and will apply to all future runs of R, regardless of where they're launched from.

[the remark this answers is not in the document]
bamader: I've added some more specific copy into the notebook along this point but tried to keep it simple to not overwhelm users.

examples/parquet_analytics_demo.ipynb
```diff
@@ -0,0 +1,476 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "d76d4d85",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Configure container access to Azure storage\n",
+    "az <- create_azure_login(tenant=[YOUR_TENANT_ID_STRING_HERE], host=[YOUR_CLOUD_HOSTED_STORAGE_PLATFORM_HERE])\n",
```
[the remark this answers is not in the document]
bamader: Mmm yep this is a good call. I'll make this change.

examples/parquet_analytics_demo.ipynb
bamader: Yeah I had that in a separate sentence but I've moved it all together to make things clearer

examples/parquet_analytics_demo.ipynb
bamader: Oh yup you're right, I did that elsewhere so I don't know why I didn't do that here

examples/parquet_analytics_demo.ipynb
bamader: I think it was your notebook that generated them but it might be corrected now

bryan-skylight: That's true that these files were created from my notebook. 🤔 A lot of the code I had in that notebook was taken from the code in `main`, though, which is why my first response was that the issue is likely there. I'll look through the notebook and see if the problem exists there. My point was that we should ensure this isn't being unintentionally done in the code somewhere before we try to "fix" it by changing the name in the storage container.

examples/parquet_analytics_demo.ipynb
```diff
@@ -0,0 +1,527 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "43223098",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "TENANT_ID = \"YOUR TENANT ID STRING\"\n",
+    "HOST = \"URL TO YOUR STORAGE PROVIDER\" # for example, https://storage.azure.com\n",
```
bryan-skylight, line 97: I think this URL needs to end with a forward slash. Can you please include that in your commented example, and maybe note it in parentheses.

examples/parquet_analytics_demo.ipynb
```diff
@@ -0,0 +1,527 @@
...
+    "ADLS_ENDPOINT = \"YOUR SPECIFIC ENDPOINT\"\n",
+    "CONTAINER = \"NAME OF YOUR STORAGE CONTAINER\""
```
bryan-skylight, line 99: Can we add the `PATH_TO_TRACT_FILES` variable here as well?

examples/parquet_analytics_demo.ipynb
```diff
@@ -0,0 +1,527 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "d50161e9",
+   "metadata": {},
+   "source": [
+    "For ease of visualization, we'll also modify the default graphics window that R uses within jupyter notebook, so that our graphs are easier to see."
```
bryan-skylight, line 107: Would it be worth pointing out to users that they can change the values based on their preferences?

bryan-skylight (approved):
Nice work!

examples/parquet_analytics_demo.ipynb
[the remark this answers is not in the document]
bryan-skylight: I think we should remove the parentheses around the text start with "This is because" and ending with "authorization issues". I think it makes sense to just "include" that in the paragraph. I leave that up to you though and think we can mark this as resolved either way.

spencer-skylight (dismissed):
LGTM!

Note: My suggestions have all been addressed except the casing for parquet file names.  It sounds like this may have more to do with the code/configuration that generated the files used. Given they were generated with those names, I think we can keep as-is for the purposes of this log.  I'll write up a quick ticket so we don't lose track of this, but I don't think this needs to be solved in this PR.

bryan-skylight (approved):

PR #145, opened 2022-06-28
Schemas.py datatype fix and performance improvements

spencer-skylight:
# Description
This PR introduces a few fixes and performance improvements described below.

## Force string datatype
Previously, iterative calls to write parquet files could fail when a single field for multiple consecutive records contained null values.  In this case, the schema derived by Pyarrow could be `string`, `double`, or some other datatype when values are present and `null` if all values in a write batch were empty.  

Updated to cast all values as strings, and store empty values as empty strings. This way, Pyarrow always derives a "string" data type.  Code will be updated in the future to support other data types, potentially by adding support for declaring data types in the schema definition.

## Hardcoded _count parameter for FHIR searches
Hardcode _count=1000 (increased from the default of 10 on Azure FHIR Server) to improve performance by increasing page size for searches, thus reducing the number of FHIR server calls.

## Cached fhirpathpy.compile calls
Previously, we called fhirpathpy.evaluate for every field in every resource, which numbers in the millions.  `fhirpathpy.evaluate` in turn called `fhirpathpy.compile` which takes about .045 seconds per call on average based on testing.

schemas.py has been updated to cache the result of fhirpathpy.compile for each FHIRPath in the schema, thus eliminating many, many calls to `fhirpathpy.compile`.  Testing shows ~10x runtime improvement with caching implemented.

### DIFF
--- a/src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
+++ b/src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
@@ -8,10 +8,11 @@ import pyarrow.parquet as pq
 import random
 import yaml
 
+from functools import cache
 from pathlib import Path
 from phdi_building_blocks.azure import AzureFhirServerCredentialManager
 from phdi_building_blocks.fhir import fhir_server_get
-from typing import Literal, List, Union
+from typing import Literal, List, Union, Callable
 
 
 def load_schema(path: str) -> dict:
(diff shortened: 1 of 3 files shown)


celestemespinoza:
Task linked: [CU-33dfnmr Optimize writing table from eCR schema](https://app.clickup.com/t/33dfnmr)

src/lib/phdi-building-blocks/phdi_building_blocks/schemas.py
```diff
@@ -63,6 +64,20 @@ def apply_selection_criteria(
...
 
+@cache
+def __get_fhirpathpy_parser(fhirpath_expression: str) -> Callable:
+    """
+    Return a fhirpathpy parser for a specific FHIRPath.  This cached function minimizes
+    calls to the relatively expensive :func:`fhirpathpy.compile` function for any given
+    `fhirpath_expression`
+
+    :param fhirpath_expression: The FHIRPath expression to evaluate
+    :return: A callable function that accepts a FHIR resource `dict` and
```
bryan-skylight, line 75: We don't include the return statement in docstrings anymore.

bamader (approved):
LGTM. I had thought based on last Friday's conversation with you, me, and Bryan that we were going to use his notebook for demo generation, but this PR looks like it puts at least some of those changes into our building blocks. Where do we stand on that aspect?

spencer-skylight:
> LGTM. I had thought based on last Friday's conversation with you, me, and Bryan that we were going to use his notebook for demo generation, but this PR looks like it puts at least some of those changes into our building blocks. Where do we stand on that aspect?

@bamader we had some back and forth on this late Friday and Monday.  Basically, the notebook code we talked about integrating ended up using quite a bit more memory than the current version in main.  After weighing the pros and cons, we decided to apply the simple workaround to the incremental update code of treating everything as a string. We have a solution design document review on the road map in the next week or so when we can decide where we want to take the architecture in the future, so we'll have a chance to incorporate the improvements in the notebook at that time.

PR #147, opened 2022-06-29
0010 FHIRPath module ADR

KennethSkylight:
Resolves [this ticket](https://reportstream-service-request.clickup.com/t/318uqev) ADR to talk about FHIRPath and the FHIRPath module used in our Python libraries


DanPaseltiner (approved):
LGTM

PR #151, opened 2022-06-30
Add Analytics ETL Refactor Prototyping Notebook

bryan-skylight:
# Description
This is the notebook I was using to toy around with various options for refactoring the analytics ETL pipeline. It looks at splitting up the `make_table` function into its three core components (extraction, tabularization, and writing), as well as how best to loop over both the schema and the resources to allow for minimal traversals and compilations of FHIR paths. Some of this work has already been incorporated into the codebase, and the rest will be discussed in the near future. We're storing this code in the `examples` directory for transparency and posterity.


bamader (approved):
LGTM. When I tried to run this locally I _once again_ hit the issue of needing to add `exclude_shared_cache_credentials=True` in the call signature of `DefaultAzureCredentials()`. I even had to copy the code for our `AzureFhirServerCredentialManager` into its own notebook cell and add that into the constructor inside the class. I'm honestly shocked no one else keeps running into this, but it's sufficiently a blocker to me that I may open a fast follows PR that just adds that into our PHDI lib class. Because our code relies on the default environment credentials, which is the last thing in the priority order that gets checked, this won't impact our functionality at all, but it  will save me (and potentially end users suffering from the same edge case) a lot of headache when it comes to why issued tokens aren't being cleared from the shared cache. But, that's not an issue with this PR.

DanPaseltiner (changes requested):
Could we have a conversation about what our standard is for merging code into Main before going forward here? I understand the importance of having this work in source control as we continue to iterate on this process, but I'm not sure that this notebook in its current form should be included in main.

PR #154, opened 2022-07-05
Remove AD data from terraform storage module

nickclyde:
This PR removes some data items from our terraform `storage` module, which attempt to read the `CT-PRIMEIngestion-AZ-Contributors` group from Active Directory in Azure, so that it can get the groups ID and set it as the owner of our `pi_datasa` storage accounts. From what I can tell, this is not really necessary, as our `SU` Azure users already have the necessary permissions to be able to interact with all storage accounts in our resource groups, regardless of owners. The reason for removing this is because CDC won't give a service principal permissions to read from Active Directory, so including this will prevent us from properly implementing CD.


celestemespinoza:
Task linked: [CU-29av3gn Create GitHub Action to Push Latest Repo Changes to Azure](https://app.clickup.com/t/29av3gn)

DanPaseltiner (approved):
LGTM

PR #159, opened 2022-07-18
Enable continuous deployment

nickclyde:
This will enable continuous deployment for the project! On every merge to main, it will trigger a deploy to the dev environment in CDC Azure. When that deploy is successful, it will trigger a deploy to the test environment. This will keep our code fully up to date in CDC Azure.

### DIFF
--- a/.github/workflows/deployDev.yaml
+++ b/.github/workflows/deployDev.yaml
@@ -1,7 +1,9 @@
 name: Deploy Dev
 
 on:
-  workflow_dispatch:
+  push:
+    branches:
+      - main
 
 jobs:
   deploy_dev:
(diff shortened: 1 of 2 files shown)


DanPaseltiner (approved):
This is awesome! Thanks for your work on this. Definitely not blocking, but do we want to automate deployment to Skylight as well? Is there a reason we aren't doing that as part of this PR?

nickclyde:
I figured that our Skylight env could act more as an environment for testing code in development, ie we could deploy to it while PRs are open for smoke testing. Adding it into the CD pipeline could cause it to get overwritten while someone else is actively using it for testing a PR

PR #161, opened 2022-07-27
New geospatial

bamader:
This changeset introduces the new SDK geospatial functionality for both raw data and fhir formatted data in accordance with the SDK PRD and SDD. GeocodeResult has been defined as a dataclass, and an abstract base class GeocodeClient has been created for each of raw and FHIR-formatted data, from which vendor-specific implementations will derive. Users are now able to geocode from strings and dictionaries (in the case of raw data) and from resources and bundles (in the case of FHIR data). Unit tests are written for every function in all new modules.

Notes:
* because `fhir` was a taken name in the directory namespace, I've temporarily called the fhir directory `fhir_wrappers`; this can change once we move everything over to the new repo
* I know the new repo is created but wasn't sure if we were ready to begin pushing code to it yet (since there's so many dependencies and things to set up in the new repo), hence this PR being opened here


spencer-skylight:
This PR has been ported to https://github.com/CDCgov/phdi-sdk

PR #162, opened 2022-07-27
Remove phdi lib

bamader:
Now that `phdi-sdk` is its own repo, the building block logic has officially migrated over. This PR removes the building blocks library from our "current" / "old" repo to account for this. Also removed are the examples/tutorials, which have migrated along with the SDK. Finally, the `requirements.txt` file now specifies the correct git path in `phdi-sdk`, which will help facilitate cleaner installations and tests when working on building blocks logic (i.e. there's no dependency on something that's in the repo necessitating separate commits and merges). 

NOTE: This PR depends on a PR Spencer will put up soon that finalizes the directory structure in `phdi-sdk` and should not be merged until the other PR lands.


src/FunctionApps/python/requirements.txt
```diff
@@ -4,7 +4,7 @@ azure-functions
 azure-identity
 azure-storage-blob
 hl7
-phdi_building_blocks @ git+https://github.com/CDCgov/prime-public-health-data-infrastructure@main#subdirectory=src/lib/phdi-building-blocks
+phdi_building_blocks @ git+https://github.com/CDCgov/phdi-sdk#subdirectory=/phdi
```
spencer-skylight, line 7: You're getting an error for this, but I think it should work if you remove the subdirectory... /phdi contains the code, but the project root (with pyproject.toml) is at the root level.

spencer-skylight (changes requested):
Just a minor suggestion for src/FunctionApps/python/requirements.txt, plus I think we need to remove the unit-test-phdi-building-blocks-lib job from /.github/workflows/test.yaml

spencer-skylight (approved):
It looks like the checks were successful, but at least for me, the PR isn't recognizing that... Maybe another commit will get it unstuck?  In any case, this looks good to me!

PR #163, opened 2022-08-19
Add ADR for Microsoft FHIR Converter

DanPaseltiner:
Adds an ADR for our decision to move forward with the Microsoft FHIR Converter.

### DIFF
--- a/docs/decisions/0013-microsoft-fhir-converter.md
+++ b/docs/decisions/0013-microsoft-fhir-converter.md
@@ -112,6 +112,6 @@ Our need for conversion of both Hl7v2 and C-CDA eliminates all options except th
 - We had success with the Microsoft FHIR Converter during our pilot with Virginia.
 - The 67 provided templates are a valuable assest.
     - These templates have met our needs so far for the conversion of ELR, VXU, and eCR messages.
-    - It is reasonable that one of the 26 templates for Hl7v2 ADT messages will meet our needs in our upcommong pilot with LA Country where we plan to begin ingesting ADT data. If none of the ADT templates are sufficient they will be usefull resources for writting our own. 
+    - It is reasonable that one of the 26 templates for Hl7v2 ADT messages will meet our needs in our upcommong pilot with LA Country, where we plan to begin ingesting ADT data. If none of the ADT templates are sufficient they will be usefull resources for writting our own. 
 - Although non-trivial, the installation process is simpler than the GCP Healthcare Data Harmonization engine and has fewer dependencies.
 - After installation, the CLI makes the tool easy to use.
--- a/docs/decisions/0013-microsoft-fhir-converter.md
+++ b/docs/decisions/0013-microsoft-fhir-converter.md
@@ -1,0 +1,105 @@
+# 12. Initial Supported File Format
+
+Date: 2022-08-19
+
+## Status
+
+Accepted
+
+## Context and Problem Statement
+
+Given that we have decided to design our buildings blocks around FHIR data, we need a solution for converting Hl7v2 and C-CDA to FHIR. Until recently, this was accomplished by using the `$convert-data` endpoint on the managed implementation of the [Microsoft FHIR server](https://github.com/microsoft/fhir-server) in Azure. However, we have begun porting our ingestion pipeline to Google Cloud Platform (GCP) where an implementation of the Microsoft FHIR server is not available. GCP refers to their implementation of a FHIR server as a [FHIR Store](https://cloud.google.com/healthcare-api/docs/reference/rest/v1/projects.locations.datasets.fhirStores) which does not have a `$convert-data` or similar endpoint. The reason for this is that a `$convert-data` endpoint is not specified as part of the [FHIR API](https://hl7.org/fhir/http.html) standard. In order to fully implement our pipeline in GCP and other platforms in the future we need a solution for FHIR conversion that can be used in any cloud environment or with on-prem infrastructure.     
+
+## Decision Drivers
+
+**Robustness** - The convert must be able to reliably and accurately convert data to FHIR. 
+
+**Hl7v2 Support** - The converter must support the conversion of Hl7v2 to FHIR. 
+
+**C-CDA Support** - The converter must support conversion of C-CDA to FHIR.
+
+**Platform Agnostic** The converter should not depend on any functionality specific to an infrastructure provider.
+
+**Ease of Implementation** - Ideally the converter will be easy implement in a variety of locations.
+
+
+## Considered Options
+
+### Microsoft FHIR Converter CLI
+
+The [Microsoft FHIR Converter](https://github.com/microsoft/FHIR-Converter) is the conversion utility behind the `$convert-data` endpoint on the Microsoft FHIR server and also offers a CLI implementation.
+
+Pros:
+- The same conversion tool we are already familiar with
+- Support for both Hl7v2 and C-CDA
+- Can be installed on Windows, MacOS, and Linux systems.
+- Open source and actively maintained by Microsoft.
+
+Cons:
+- Installation requires building from source.
+- Written in C# with .NET framework making integration with PHDI's primarily Python based work challenging.
+
+### LinuxForHealth FHIR Converter
+
+The [LinuxForHealth FHIR Converter](https://github.com/LinuxForHealth/hl7v2-fhir-converter) is an open source Hl7v2 to FHIR conversion utility written in Java.
+
+Pros:
+- Supports Hl7v2
+- Open source 
+- Can be built on any system with a JVM.
+
+Cons:
+- Does not support C-CDA
+- Installation process is non-trivial
+
+### Report Stream Fhirengine
+
+The Report Stream project is developing a FHIR conversion tool called [fhirengine](https://github.com/CDCgov/prime-reportstream/tree/master/prime-router/src/main/kotlin/fhirengine). It is open source, written in Kotlin, and supports bi-directional Hl7v2 and FHIR conversion.
+
+Pros:
+- Supports Hl7v2
+- Supports conversion from FHIR back to Hl7v2.
+- Can be built on any system with a JVM.
+- Designed, built, and maintained by another project with in PRIME.
+
+Cons:
+- Does not support C-CDA.
+- Installation process is non-trivial.
+ 
+
+### CDC eCR Team FHIR Converter
+
+The Electronic Case Reporting team at CDC is developing a FHIR conversion tool.
+
+Pros:
+- Supports C-CDA
+- Developed by another team at CDC.
+
+Cons:
+- We believe that it does not support Hl7v2, but have not be able to confirm this.
+- The eCR team has not be responsive to setting up a meeting to dsicuss their converter.
+
+### GCP Healthcare Data Harmonization
+
+The [GCP Healthcare Data Harmonization](https://github.com/GoogleCloudPlatform/healthcare-data-harmonization) engine offers tools for conversion from Hl7v2 and C-CDA to FHIR.
+
+Pros:
+- Supports Hlv2
+- Supports C-CDA
+- Open source
+
+Cons:
+- Documentation and examples do not seem to cover C-CDA conversion although reviewing the code base shows that C-CDA conversion is supported.
+- Installation has several dependencies and is non-trivial
+
+
+## Decision Outcome
+
+**Microsoft FHIR Converter** 
+
+Our need for conversion of both Hl7v2 and C-CDA eliminates all options except the Microsoft FHIR Converter and GCP Healthcare Data Harmonization engine. We have elected to move forward with Microsoft FHIR Converter for the following reasons:
+
+- As a team we are already familiar with the Microsoft FHIR Converter.
+- Although non-trivial the installation process is simpler than the GCP Healthcare Data Harmonization engine and has fewer dependencies.
+- After installation the CLI makes the tool easy to use.
+- We have been able to build a containerized application around this converter that makes its functionality availabe via an HTTP API and can be run on any platform that supports Docker containers.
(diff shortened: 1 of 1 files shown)


celestemespinoza:
Task linked: [CU-2t3tzna SPIKE: Investigation FHIR conversion methods in GCP](https://app.clickup.com/t/2t3tzna)

docs/decisions/0013-microsoft-fhir-converter.md
```diff
@@ -0,0 +1,105 @@
...
+
+Given that we have decided to design our buildings blocks around FHIR data, we need a solution for converting Hl7v2 and C-CDA to FHIR. Until recently, this was accomplished by using the `$convert-data` endpoint on the managed implementation of the [Microsoft FHIR server](https://github.com/microsoft/fhir-server) in Azure. However, we have begun porting our ingestion pipeline to Google Cloud Platform (GCP) where an implementation of the Microsoft FHIR server is not available. GCP refers to their implementation of a FHIR server as a [FHIR Store](https://cloud.google.com/healthcare-api/docs/reference/rest/v1/projects.locations.datasets.fhirStores) which does not have a `$convert-data` or similar endpoint. The reason for this is that a `$convert-data` endpoint is not specified as part of the [FHIR API](https://hl7.org/fhir/http.html) standard. In order to fully implement our pipeline in GCP and other platforms in the future we need a solution for FHIR conversion that can be used in any cloud environment or with on-prem infrastructure.     
+
+## Decision Drivers
+
+**Robustness** - The convert must be able to reliably and accurately convert data to FHIR.
```
bryan-skylight: Typo: "convert" -> "converter"

We should explicitly state that it should be as lossless as possible.

docs/decisions/0013-microsoft-fhir-converter.md
```diff
@@ -0,0 +1,105 @@
...
+**Robustness** - The convert must be able to reliably and accurately convert data to FHIR. 
+
+**Hl7v2 Support** - The converter must support the conversion of Hl7v2 to FHIR. 
+
+**C-CDA Support** - The converter must support conversion of C-CDA to FHIR.
+
+**Platform Agnostic** The converter should not depend on any functionality specific to an infrastructure provider.
+
+**Ease of Implementation** - Ideally the converter will be easy implement in a variety of locations.
+
```
bryan-skylight, line 26: We should add Extensibility as a decision driver and point out that MSFT's ability to accommodate custom templates means any deviations from base HL7v2 specifications can be handled.

docs/decisions/0013-microsoft-fhir-converter.md
```diff
@@ -0,0 +1,105 @@
...
+**Robustness** - The convert must be able to reliably and accurately convert data to FHIR. 
+
+**Hl7v2 Support** - The converter must support the conversion of Hl7v2 to FHIR. 
+
+**C-CDA Support** - The converter must support conversion of C-CDA to FHIR.
+
+**Platform Agnostic** The converter should not depend on any functionality specific to an infrastructure provider.
+
+**Ease of Implementation** - Ideally the converter will be easy implement in a variety of locations.
```
bryan-skylight: Missing word: "will be easy implement" -> "will be easy to implement"

docs/decisions/0013-microsoft-fhir-converter.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Decision Drivers
+
+**Robustness** - The convert must be able to reliably and accurately convert data to FHIR. 
+
+**Hl7v2 Support** - The converter must support the conversion of Hl7v2 to FHIR. 
+
+**C-CDA Support** - The converter must support conversion of C-CDA to FHIR.
+
+**Platform Agnostic** The converter should not depend on any functionality specific to an infrastructure provider.
```
bryan-skylight: nit: Can we reword this to be more clear as to what it's saying? Maybe something like "The conversion process should not depend on a specific cloud environment's implementation."?

docs/decisions/0013-microsoft-fhir-converter.md
```diff
@@ -0,0 +1,105 @@
...
+
+
+## Considered Options
+
+### Microsoft FHIR Converter CLI
+
+The [Microsoft FHIR Converter](https://github.com/microsoft/FHIR-Converter) is the conversion utility behind the `$convert-data` endpoint on the Microsoft FHIR server and also offers a CLI implementation.
+
+Pros:
+- The same conversion tool we are already familiar with
```
bryan-skylight: Recommendation: Change this to something like "The same conversion tool that we had success with in our pilot with VA". The reason I like this over what you have is that it focuses more on impact (we had success with it) and less on our personal preference (it's what we're familiar with).

docs/decisions/0013-microsoft-fhir-converter.md
```diff
@@ -0,0 +1,105 @@
...
+- Support for both Hl7v2 and C-CDA
+- Can be installed on Windows, MacOS, and Linux systems.
+- Open source and actively maintained by Microsoft.
+
+Cons:
+- Installation requires building from source.
```
bryan-skylight: This isn't necessarily true, right? Couldn't we pre-build the package and provide the files in a zip file that users download? Running from the command line is definitely a con, though, as it still requires .NET be installed for non-Windows users.

docs/decisions/0013-microsoft-fhir-converter.md
```diff
@@ -0,0 +1,105 @@
...
+Cons:
+- Does not support C-CDA
+- Installation process is non-trivial
+
+### Report Stream Fhirengine
+
+The Report Stream project is developing a FHIR conversion tool called [fhirengine](https://github.com/CDCgov/prime-reportstream/tree/master/prime-router/src/main/kotlin/fhirengine). It is open source, written in Kotlin, and supports bi-directional Hl7v2 and FHIR conversion.
+
+Pros:
+- Supports Hl7v2
```
bryan-skylight: Currently only supports Hl7v2 ORU_R01 messages. It doesn't support the full HL7v2 standard.

docs/decisions/0013-microsoft-fhir-converter.md
```diff
@@ -0,0 +1,105 @@
...
+- Supports conversion from FHIR back to Hl7v2.
+- Can be built on any system with a JVM.
+- Designed, built, and maintained by another project with in PRIME.
```
bryan-skylight: typo: "with in" -> "within"

docs/decisions/0013-microsoft-fhir-converter.md
```diff
@@ -0,0 +1,105 @@
...
+### CDC eCR Team FHIR Converter
+
+The Electronic Case Reporting team at CDC is developing a FHIR conversion tool.
+
+Pros:
+- Supports C-CDA
+- Developed by another team at CDC.
+
+Cons:
+- We believe that it does not support Hl7v2, but have not be able to confirm this.
```
bryan-skylight: I think we can definitively state that it doesn't support Hl7v2 based on comments from Rick Hawes during our of meetings.

docs/decisions/0013-microsoft-fhir-converter.md
```diff
@@ -0,0 +1,105 @@
...
+- The eCR team has not be responsive to setting up a meeting to dsicuss their converter.
```
bryan-skylight: Let's rephrase this to be less accusatory. Perhaps something like "The source code for this tool is not readily available, and the documentation is sparse, making it hard to assess the tool's suitability"?

docs/decisions/0013-microsoft-fhir-converter.md
```diff
@@ -0,0 +1,105 @@
...
+
+
+## Decision Outcome
+
+**Microsoft FHIR Converter** 
+
+Our need for conversion of both Hl7v2 and C-CDA eliminates all options except the Microsoft FHIR Converter and GCP Healthcare Data Harmonization engine. We have elected to move forward with Microsoft FHIR Converter for the following reasons:
+
+- As a team we are already familiar with the Microsoft FHIR Converter.
+- Although non-trivial the installation process is simpler than the GCP Healthcare Data Harmonization engine and has fewer dependencies.
```
bryan-skylight: nit: "Although non-trivial the installation" -> "Although non-trivial, the installation"

docs/decisions/0013-microsoft-fhir-converter.md
```diff
@@ -0,0 +1,105 @@
...
+- After installation the CLI makes the tool easy to use.
```
bryan-skylight: nit: "After installation the CLI" -> "After installation, the CLI"

docs/decisions/0013-microsoft-fhir-converter.md
```diff
@@ -0,0 +1,105 @@
...
+- We have been able to build a containerized application around this converter that makes its functionality availabe via an HTTP API and can be run on any platform that supports Docker containers.
```
bryan-skylight: We shouldn't list this as part of our decision factors. We very likely could have done the same with Google's option. I would replace this with a bullet point highlighting that Microsoft's solution comes out of the box with 57 Hl7v2 templates and 9 C-CDA templates.

docs/decisions/0013-microsoft-fhir-converter.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Considered Options
+
+### Microsoft FHIR Converter CLI
+
+The [Microsoft FHIR Converter](https://github.com/microsoft/FHIR-Converter) is the conversion utility behind the `$convert-data` endpoint on the Microsoft FHIR server and also offers a CLI implementation.
+
+Pros:
+- The same conversion tool we are already familiar with
+- Support for both Hl7v2 and C-CDA
```
bryan-skylight: We should also not that it supports "upscaling" FHIR from STU3 to R4, and will likely support R4 to R5 in the future when R5 comes out.

docs/decisions/0013-microsoft-fhir-converter.md
```diff
@@ -0,0 +1,105 @@
...
+
+Pros:
+- Supports C-CDA
+- Developed by another team at CDC.
+
+Cons:
+- We believe that it does not support Hl7v2, but have not be able to confirm this.
+- The eCR team has not be responsive to setting up a meeting to dsicuss their converter.
+
+### GCP Healthcare Data Harmonization
```
bryan-skylight, line 89: We should be more clear in our description of this option. Based on a brief read of the documentation, the tool appears to be a general purpose converter from one data format to another. Some cons that jump out to me are that there don't appear to be any pre-built templates, meaning we'd either have to create them ourselves or have the STLTs do it, and running the code outside of Google's cloud environment seems to require having Go installed. You've mentioned some of this below, but I think we can make it a little more explicit.

bryan-skylight (changes requested):
Nice work! I've got some nit-picky things, as well as some recommendations that I'd like to see considered before we merge this.

docs/decisions/0013-microsoft-fhir-converter.md
DanPaseltiner: Agreed.

docs/decisions/0013-microsoft-fhir-converter.md
DanPaseltiner: Agreed on STU3 to R4. How do we know that it will "likely support R4 to R5". My thought is that this was really just our own speculation, but not something we actually know. We do know that the MSFT converter is actively maintained and that MSFT is committed to expanding further into healthcare. Maybe getting a little pedantic here.

docs/decisions/0013-microsoft-fhir-converter.md
DanPaseltiner: It is true in terms of what MSFT provides. We could definitely provide a pre-built version as you describe, but I would argue that in that case PHDI is getting involved in the installation process. A STLT would install the PHDI pre-built MSFT FHIR Converter, not the MSFT Converter directly.

docs/decisions/0013-microsoft-fhir-converter.md
DanPaseltiner: Good point. Until you mentioned this the other day, I didn't really appreciate how limited the tool currently is.

docs/decisions/0013-microsoft-fhir-converter.md
DanPaseltiner: Good call out.

docs/decisions/0013-microsoft-fhir-converter.md
bryan-skylight: I would argue that providing the pre-built software would be akin to what we're doing with the Smarty API: we're providing a wrapper that makes it easier for STLTs to use. I'm not saying that's what we should do, and I'm happy to keep the language as is. My primary point was more around the use of the word "requires" as that's not technically true. It just depends on how much we want to help facilitate the installation.

docs/decisions/0013-microsoft-fhir-converter.md
bryan-skylight: That's a fair point; we probably shouldn't put speculation in the ADR.

bryan-skylight (dismissed):
Your counterpoints are valid, so I'm good leaving the language as is in those two instances. Thanks for putting this together!

docs/decisions/0013-microsoft-fhir-converter.md
DanPaseltiner: Gotcha. How about "As provided by Microsoft, installation requires building from source"?

bryan-skylight:
The `unit-test-python` check is failing due to some breaking changes we made to this repository. While we could fix this issue, this repository is, for all intents and purposes, deprecated and won't be getting the attention it needs for these things. With that said, I'm opting to manually override these checks to let this PR land.

bryan-skylight (approved):

